Issue #22, opened 2021-04-20
Connect stream player to arcsi api


baxgas:
connect show title and link to show. refine buttons

PR #56, opened 2021-05-06
#22 show page link in player

gammaw:
A quick solution for the show URL in the player. Not sure how good my Vue/Nuxt design decisions were but it seems to work. Anyway, it was fun to get to now the basics of Nuxt and do something with it!

### DIFF
--- a/frontend/components/HeaderBlock.vue
+++ b/frontend/components/HeaderBlock.vue
@@ -2,7 +2,7 @@
   <header class="container flex-row-reverse py-4 md:flex md:flex-row">
     <div>
       <client-only>
-        <RadioPlayer :show_album_art="true" :now_playing_uri="streamServer" />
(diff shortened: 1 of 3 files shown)


baxgas (commented):
nice you put the font awesome icons.
it will depend on the design, but generally I think these icons are nicer and fresher design: https://boxicons.com/
also, I would unify the icons, since now it is material icons that is used at the play/volume and font awesome for the link and I think at Wordpress page at other places. @gammaw — can I go with it?

baxgas (approved):
@gammaw fantastic work!!! It can go in master. 
or if you want to do also the defaultshow image it is also quite easy with copy/paste and endpoint name adjustments to the code...

gammaw:
Thanks @baxgas! As we discussed we'll do the artwork issue separately.

Issue #58, opened 2021-05-10
Add fallback show image in radio player


baxgas:
If there is no image in the stream player, use the one from arcsi

PR #59, opened 2021-05-10
#58 Fallback image for stream

baxgas:
@gammaw – I made the stream player fallback to the show image, please check if it follows the logic of the old one for the live shows eg.
Also fixed a bit the caching of the store object, so that it reads out the current show only once, and that object is referenced in the scripts. I had an edge case to fix with Erdenklang ndash and dash difference, so I implemented a slugify script to compare the titles, which should work also for Svadhyaya, dodo and other titles with unusual characters

### DIFF
--- a/components/RadioPlayer.vue
+++ b/components/RadioPlayer.vue
@@ -188,7 +188,6 @@ export default {
       default_azuracast_art_url: 'https://streaming.lahmacun.hu/static/img/generic_song.jpg',
       showsURLList_lookup: [],
       showsList_lookup: [],
-      arcsiShows: this.$store.state.arcsiShows
     }
   },
   computed: {
@@ -207,6 +206,16 @@ export default {
         })
       })
       return allStreams
+    },
+    arcsiList () {
+      return [...this.$store.state.arcsiShows]
+    },
+    currentShowArcsi () {
+      if (this.np.live.is_live) { // live show
+        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.live.streamer_name))
+      } else { // pre-recorded show
+        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.now_playing.song.artist))
+      }
     },
     time_percent () {
       const timePlayed = this.np.now_playing.elapsed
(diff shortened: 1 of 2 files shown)


components/RadioPlayer.vue
```diff
@@ -208,6 +207,16 @@ export default {
...
+    arcsiList () {
+      return [...this.$store.state.arcsiShows]
+    },
+    currentShowArcsi () {
+      if (this.np.live.is_live) { // live show
+        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.live.streamer_name))
+      } else { // pre-recorded show
+        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.now_playing.song.artist))
+      }
+    },
```
gammaw, line 219: The basic logic looks good to me. 

Two questions:
- Is it really slugifying? It rather looks to me to be a normalization of special characters so that the strings are comparable. 
- What will this function return if there is no hit?

components/RadioPlayer.vue
```diff
@@ -255,34 +264,20 @@ export default {
...
-      if (this.np.live.is_live) // live show
-      { this_show = arcsiShowsList.find(show => show.name === this.np.live.streamer_name) } else // pre-recorded show
-      { this_show = arcsiShowsList.find(show => show.name === this.np.now_playing.song.artist) }
-      let url = ''
-      try {
-        url = this_show.archive_lahmastore_base_url
-      } catch (error) { // Happens on the first call after page is loaded, not sure why...
-        // console.log("No show URL found.";
-      }
+      const url = this.currentShowArcsi ? this.currentShowArcsi.archive_lahmastore_base_url : ''
```
gammaw, line 267: Will this be a false branch if currentShowArcsi() has no hit (see above)?

components/RadioPlayer.vue
```diff
@@ -255,34 +264,20 @@ export default {
...
+        return this.currentShowArcsi ? this.currentShowArcsi.cover_image_url : this.default_art_url
       } else {
         const songTitleJSON = this.np.now_playing.song.title
         const songArtistJSON = this.np.now_playing.song.artist
         const artworkJSON = this.np.now_playing.song.art // art work url in json
 
         if (artworkJSON === this.default_azuracast_art_url) { // default url by azuracast (must be returning off air music with art work)
-          const tryArtFromShow = this.showsList_lookup[songArtistJSON] // try to find show artwork url based on show title
-          if (tryArtFromShow === undefined) { // show not found
+          if (this.currentShowArcsi.cover_image_url === undefined) { // show not found
```
gammaw, line 280: Let's get rid of searching through the history, no? I don't think it's used anyways. 

Plus, is it a real threat that `this.currentShowArcsi.cover_image_url` is undefined? It can be an assumption on arcsi's side, no?

components/RadioPlayer.vue
baxgas: slugifying is because the disrepancies between html charactes in names eg. Erdenklang, or Svadhyaya
it will return null, do you think we need some fallback? but to what?

components/RadioPlayer.vue
baxgas: it will be empty string if this.currentShowArcsi is false/null/undefined

components/RadioPlayer.vue
baxgas: I don't know why it was checked, but it was in the old script, maybe there were some cases like that? I don't think it is bad to check for it, if it's ok, it will just pass it, no?

Issue #57, opened 2021-05-08
Make sub-pages for Arcsi Episodes


baxgas:
Every Arcsi Episode is a new route

gammaw:
I guess we'll need to find a generic solution for the CORS issue (see https://github.com/mmmnmnm/lahmacun_rework/pull/60). Browser plugins are local solution and won't fix the issue on the user side, at least it still doesn't work on my end.

PR #60, opened 2021-05-18
#57 Arcsi page implement

baxgas:
I implemented the Arcsi sub-page, now with a different method to fetch the exact episode by ID.
TODO - instead of ID to have a slug of the episode?

### DIFF
--- a/constants/index.js
+++ b/constants/index.js
@@ -1,4 +1,5 @@
 export const arcsiServerURL = 'https://arcsi.lahmacun.hu/arcsi/show/all'
+export const arcsiItemBaseURL = 'https://arcsi.lahmacun.hu/arcsi/item'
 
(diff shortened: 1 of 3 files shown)


gammaw (commented):
Nice to see the per-item sub pages!

I noticed something weird: when I visit an item page from a show page, then I get a CORS error and no item data is shown (in Chrome, running on localhost). When I refresh the URL of the item page, then the page loads all right. I couldn't figure out what the cause of it is, but I wasn't looking too long. Can you reproduce it? 

I also noticed that you use another way to fetch the item API (not via store, not via axios). Does it have a particular reason?

baxgas:
I managed to overcome CORS with a Firefox plugin, otherwise it doesn't work
(Chrome eg.)

The fetch is dyamic per episode item, via it's id. This fetch method gets
also cached server-side, so it is the prefered way in nuxt components

Issue #64, opened 2021-05-22
Make homepage according to Csenge's design


baxgas:
![image](https://user-images.githubusercontent.com/36408851/119220268-0261a780-baea-11eb-8476-c3de91125f85.png)

PR #65, opened 2021-06-03
#64 new design header layout

baxgas:
Made the home page layout according to Csenge's Designs, not all parts are finished, but if you approve it, this could be demoed to the UX/UI team


pages/index.vue
```diff
@@ -1,23 +1,10 @@
...
+  <div>
+    <section class="grid-cols-2 mb-16 md:grid">
       <div>
-        <nuxt-link to="news/">
-          <h3>News</h3>
-        </nuxt-link>
-        <div v-for="news in sortNews" :key="news.id">
-          <NewsBlock :news="news" />
-        </div>
+        <NewsHome :news="sortNews[newsStart]" @changenews="changenews" />
```
gammaw, line 5: Wow, didn't know that you can refer to the /news/Home.vue component like this. 
(I found it also in the docs: https://nuxtjs.org/docs/2.x/directory-structure/components/)

gammaw (commented):
Seems to be a nice structure! I guess I got the big picture. 
Added a question and comment. 

Thanks a lot for the this big chunk of work!

components/HeaderBlock.vue
```diff
@@ -1,13 +1,20 @@
 <template>
-  <header class="container flex-row-reverse py-4 md:flex md:flex-row">
-    <div>
-      <client-only>
-        <RadioPlayer :show-album-art="true" :now-playing-uri="streamServer" />
-      </client-only>
+  <header class="flex-row-reverse items-center justify-between px-4 py-4 bg-white md:flex md:flex-row">
```
pvj, line 2: Wow! I learned a lot looking these classes up.

components/HeaderBlock.vue
```diff
@@ -1,13 +1,20 @@
...
+    <div class="flex items-center">
+      <div class="block mr-4 cursor-pointer" @click="$router.push('/')">
+        <img src="/img/lahmacun-logo.png" alt="" class="w-24">
+      </div>
+      <div class="mr-8">
+        <client-only>
+          <RadioPlayer :show-album-art="true" :now-playing-uri="streamServer" />
```
pvj, line 9: The `uri` part is really cool. Does this set-up mean we could - technically - easily use the same player for `streamServer` and then on click for `arcsi` and then back on streamServer. 

Maybe this is not the right place to discuss it, commenting on 1 line of code in the middle of PR but I'm thinking lot about the following: 

Let's say we have only 1 unified player. By default it plays the broadcast stream. User arrives and clicks on an arcsived episode to listen. This is fairly clear where the button is to listen to an old episode. The player switches from broadcast to playback. Here's my question: How/where can you switch back to listen to the radio? I've been thinking about this for some time and this is not clear to me.

components/HeaderBlock.vue
```diff
@@ -1,13 +1,20 @@
...
+        </client-only>
+      </div>
     </div>
-    <div class="flex items-center cursor-pointer" @click="$router.push('/')">
-      <img src="/img/lahmacun-logo.png" alt="" class="h-16 mr-4">
-      <h1>Lahmacun Radio</h1>
+    <HeadNavigation />
+    <div id="time">
+      <div class="currenttime">
+        {{ timestamp }}
```
pvj, line 16: I know you are going by the design but would it be possible to add two timezones? One is your computer's time one is the lahmacun time (UTC+2).

components/HeaderBlock.vue
```diff
@@ -22,15 +29,25 @@ export default {
...
-      streamServer
+      streamServer,
+      timestamp: ''
     }
   },
   computed: {
 
   },
+  created () {
+    setInterval(this.getNow, 1000)
```
pvj, line 40: nice!

components/HeaderBlock.vue
```diff
@@ -22,15 +29,25 @@ export default {
...
+  },
+
   mounted () {
   },
   methods: {
+    getNow () {
```
pvj, line 46: Question for the future: 
the `getNow()` function might be global function called on different multiple components, should we move it to some other place? 

Perhaps we will add /schedule page where we display crosshair to show the user where we are in the program. That crosshair could use the exact same method (`getNow()`) and exact same logic (`setInterval(this.getNow, 1000)`)

components/RadioPlayer.vue
```diff
@@ -7,6 +7,14 @@
 
       <div class="now-playing-details">
         <div class="radio-controls">
+          <div v-if="showAlbumArt && np.now_playing.song.art" class="now-playing-art">
```
pvj, line 10: is `showAlbumArt` settable somewhere? Could we maybe create a config for the player with list of such options? 

- show album art
- display elapsed time
- display remaining time
- display chatbox
- display `ON AIR` banner

etc. 

That could be really useful for all the different types of things we will be using the player for later (eg. live shows, repeats, shared broadcasts, one-off events etc.)

components/RadioPlayer.vue
```diff
@@ -7,6 +7,14 @@
...
+            <a class="cursor-pointer programimage" rel="playerimg" @click.stop="streamModal = !streamModal">
+              <div v-if="show_check == true" class="onair">On air</div>
+              <img class="progimg" :src="show_art_url" :alt="'album_art_alt'">
+            </a>
+            <Modal :media="show_art_url" :title="show_title" :description="show_subtitle" :visibility="streamModal" @close="closeModal" />
```
pvj, line 15: What happens on the `@close` action?

components/RadioPlayer.vue
```diff
@@ -7,6 +7,14 @@
...
+          </div>
+
           <div class="play-volume-controls">
```
pvj, line 18: This is again maybe the wrong place -- let me know if we should move this discussion just say please create a separate issue for this, dude and I will do so -- 

Do you think we should add skip-backward skip-forward buttons? Especially for arcsi items that would be useful to be able to use them instead of seeking by hand

components/RadioPlayer.vue
```diff
@@ -22,41 +30,33 @@
             <a href="#" class="text-secondary volumeshower" @mouseenter="showVolumeSlider = !showVolumeSlider">
```
pvj, line 30: How come these are not part of the player modal? Maybe I'm misunderstanding what the Modal here means or what the role of the Modal is? Can  you link me to some place where I may dive deeper?

components/RadioPlayer.vue
```diff
@@ -22,41 +30,33 @@
...
-                step="1"
-              >
-            </div>
-            <div class="radio-control-max-volume-button">
-              <a href="#" class="text-secondary" :title="'full_volume_btn'" @click.prevent="volume = 100">
-                <i class="material-icons" aria-hidden="true">volume_up</i>
-              </a>
+            <div v-show="showVolumeSlider" id="radio-player-controls" class="radio-controls-standalone volumecontrolos" @mouseleave="showVolumeSlider = !showVolumeSlider">
+              <div class="radio-control-mute-button">
+                <a href="#" class="text-secondary" :title="'mute_btn'" @click.prevent="volume = 0">
```
pvj, line 36: Does here `@click.prevent` event mean that it prevents clicking _if_ the volume is already 0? 

Great that we are adding mute functionality! This was 1 thing that people asked about a lot!!

components/HeaderBlock.vue
```diff
@@ -1,13 +1,20 @@
...
+        <client-only>
+          <RadioPlayer :show-album-art="true" :now-playing-uri="streamServer" />
+        </client-only>
+      </div>
     </div>
-    <div class="flex items-center cursor-pointer" @click="$router.push('/')">
-      <img src="/img/lahmacun-logo.png" alt="" class="h-16 mr-4">
-      <h1>Lahmacun Radio</h1>
+    <HeadNavigation />
+    <div id="time">
```
pvj, line 14: I like your logic of having HeaderBlock > HeaderNavigation 

Is there a reason time is not a component of its own?

components/ScheduleHome.vue
```diff
@@ -1,15 +1,26 @@
 <template>
```
pvj, line 1: This is just me thinking out loud maybe this is stupid thing to ask or assume in Vue-world. 

I know we display this schedule on the home page. But I believe the template will be the same on the schedule page daily view as well.

So do you think we could re-use this component there as well? 

Because if we can I would rename the component to something like: ScheduleDailyList

And then this component can be used on `/` and also on `/schedule/list` too without any modification? At least this is my dream-like scenario. Then even if you switch the days on `/schedule/list` page from today to tomorrow or 2028. 11. 04. we would use this component for rendering that as well just w/ different data.

And you could just add `ScheduleWeeklyList` or `ScheduleDailyGrid` or `ScheduleWeeklyGrid` etc. for the different displays and designs -- maybe there is some `ScheduleAbstract` component that is used by all these more custom components (I'm not sure I know enough about inheritance in Vue-world).

Again, sorry if this is the wrong train of thought but I wanted to mention these ideas.

components/ScheduleHome.vue
```diff
@@ -1,15 +1,26 @@
...
       </div>
-      <div class="dayshows">
-        <div v-for="show in day" :key="show.id" :class="showAirCheck(index, show.name) ? 'dayblock onair' : 'dayblock'">
+      <div class="infos day">
+        <h3>{{ todayName }}</h3>
+      </div>
+    </div>
+    <div class="flex items-center scheduleblock">
+      <div class="w-full m-auto">
+        <div v-for="show in showsByDate[0]" :key="show.id" :class="showAirCheck(0, show.name) ? 'dayblock onair' : 'dayblock'">
```
pvj, line 15: Maybe later on we should add a nice error message if there is no data?

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
...
     </div>
+    <div v-if="false" class="old-schedule">
+      <div v-for="(day, index) in showsByDate" :key="day.id" class="mb-2 scheduleblock">
+        <div class="flex flex-row justify-between dayname">
+          <h4>{{ dayNames[day[0].day-1] }}</h4>
+          <span>{{ format(new Date(index === 0 ? todayDate :
+            tommorrow.setDate(tommorrow.getDate() + 1)) , 'MMMM dd.') }}</span>
+        </div>
+        <div class="dayshows">
+          <div v-for="show in day" :key="show.id" :class="showAirCheck(index, show.name) ? 'dayblock onair' : 'dayblock'">
```
pvj, line 43: This is an awesome solution! So cool that you can evaluate this `showAirCheck()` function for deciding which class to use

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
...
+            <div class="show-basic-infos">
+              {{ removeSeconds(show.start) }}-{{ removeSeconds(show.end) }} -
```
pvj, line 45: Cool!

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
...
+              <nuxt-link :to="'/shows/' + show.archive_lahmastore_base_url">
```
pvj, line 46: Wow!! Cool!!! This is honestly cool to see applied like this

pvj: Makes me think I should've found a better name for this field

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
...
+                {{ show.name }}
+              </nuxt-link>
+              <div class="show-image" :style="{ backgroundImage: `url(${show.cover_image_url})` }" />
```
pvj, line 49: Wow! Awesome! I was just thinking about adding the show cover image for each show in the schedule

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
...
+            </div>
+            <div v-if="showAirCheck(index, show.name)" class="onairtext">
```
pvj, line 51: Maybe if the show is on air we could bold the text as well? Or display it w/ bigger font? Or would that be too much of an effect? Maybe it is better to be subtle. Well anyway this is definitely not for me to decide

components/ScheduleHome.vue
```diff
@@ -64,6 +98,9 @@ export default {
       } else {
         return this.nowPlaying?.now_playing?.song.artist
```
pvj, line 99: This conditional operator is actually pretty cool!  

But if it returns none then how do we handle it? Since it is already in the `else` statement part of execution

components/news/Home.vue
```diff
@@ -0,0 +1,115 @@
...
+      <div class="news-date">
+        <div class="newsgeneral">
+          <nuxt-link to="news/">
+            <h3 class="m-0">
+              News
+            </h3>
+          </nuxt-link>
+        </div>
+        <div class="newsdate">
+          <p>{{ format(new Date(news.modified), 'MM') }}</p>
```
pvj, line 18: It seems like we are using time and date in many places. 

When you develop in js / vue what's preferred? To always just format these and create time / date objects inplace or to have generic methods somewhere and use those for all over the application? 

Sorry if this is dumb question but I don't know what the best practices are for current front-end development.

components/news/Home.vue
```diff
@@ -0,0 +1,115 @@
...
+          <p>{{ format(new Date(news.modified), 'dd') }}</p>
+        </div>
+      </div>
+      <div class="px-4 news-info-details">
+        <h5 class="py-4">
+          <nuxt-link :to="'/news/' + news.slug">
+            {{ htmlDecoder(news.title.rendered) }}
```
pvj, line 25: I searched our repo for htmlDecoder this is 1000% amazing & smart mixin method! https://github.com/mmmnmnm/lahmacun_rework/blob/03990598f3cee4655f13c949bedfbdb55f74f6ed/plugins/mixinCommonMethods.js#L34 I think you were quite smart to handle it this way

pvj: Maybe these are not related. 

So I searched for the `htmlDecoder()` stuff and I found that you have a `news/_slug` object and `head()` method there returning the title already what is the difference between that and this, why don't we just use that?

components/news/Home.vue
```diff
@@ -0,0 +1,115 @@
...
+          </nuxt-link>
+        </h5>
+        <div v-html="news.excerpt.rendered" />
+        <div class="flex justify-end my-4">
+          <div class="px-4 py-2 bg-gray-200 rounded-lg cursor-pointer more-button" @click="$emit('changenews')">
```
pvj, line 30: Well I was never fan and never understood the js stuff but I have to say these click events and this emit stuff is really really smooth & cool. Good job!

components/news/Home.vue
```diff
@@ -0,0 +1,115 @@
...
+  },
+  watch: {
+    news: {
+      handler () {
+        this.loadNewsImages(this.news.featured_media)
+      }
+
+    }
+  },
+  mounted () {
```
pvj, line 68: Is it okay that `handler()` is part of the `news:` obj while `mounted()` is part of `watch:` 

Can you link me the docs or some place to understand this area of Vue (js?) better?

components/news/Home.vue
```diff
@@ -0,0 +1,115 @@
...
+    this.loadNewsImages(this.news.featured_media)
+  },
+  methods: {
+    async loadNewsImages (newsId) {
+      const adress = `${contentApiURL}/media/${newsId}`
+      const responseNews = await this.$axios.get(adress)
+      this.newsImage = responseNews.data?.media_details?.sizes?.large?.source_url || responseNews.data?.source_url || this.newsImage
```
pvj, line 75: Cool! This means we can optimise images for example to save bandwith on mobile networks?

layouts/default.vue
```diff
@@ -1,8 +1,7 @@
 <template>
   <div>
     <HeaderBlock />
-    <head-navigation />
-    <nuxt />
+    <nuxt class="py-8" />
```
pvj, line 4: python? :D Just kidding but honestly I'm curious what is py here?

pages/index.vue
```diff
@@ -1,23 +1,10 @@
 <template>
```
pvj, line 1: Can you draw a tree or explain with `>` `<` relation marks how to think about `layouts/` vs. `pages/`? 

I see we have `default` and `home` layouts and I think I understand that those are the templates and then `pages/` are the actual views where you combine templates w/ data?

pages/index.vue
```diff
@@ -26,15 +13,13 @@
 
 <script>
 export default {
-  components: {
-  },
+  layout: 'home',
   data () {
     return {
-      newsLimit: 3
+      newsLimit: 9,
```
pvj, line 19: Nice!

pages/index.vue
```diff
@@ -53,14 +38,19 @@ export default {
     }
   },
   methods: {
+    changenews () {
```
pvj, line 41: nice method! 

Why do you subtract 1 why don't we let `newsStart` become the same number as `newsLimit`? Is this an index starting from 0?

pvj (approved):
I approve based on the results this is tremendous boost to our development efforts. I don't even know how you had the time but anyway you did it. 

Great job Gas! One of thebiggest changesets I had to review in quite some time and in language / framework / discipline I don't know much about. Sorry if I made your life harder!

components/ScheduleHome.vue
```diff
@@ -21,6 +32,29 @@
         </div>
       </div>
     </div>
+    <div v-if="false" class="old-schedule">
```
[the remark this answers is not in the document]
baxgas: I just left the code for the old schedule invisible, I think we'll need some chunks from it eg. to show the other days (other than current)

pages/index.vue
baxgas: yes, folder/names can be used also

assets/css/_variables.scss
```diff
@@ -0,0 +1,3 @@
+$lahma-pink: #d09cf8;
+$black-color: #23282d;
+$lahma-type: 'Rubik', "Segoe UI","Helvetica Neue",Arial,sans-serif,"Apple Color Emoji","Segoe UI Emoji","Segoe UI Symbol"
```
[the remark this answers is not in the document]
baxgas: She used the same font and lahma-pink, but sure, she can decide about this

components/FooterHome.vue
```diff
@@ -0,0 +1,93 @@
+<template>
```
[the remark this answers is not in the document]
baxgas: for now I separated the layout of the home page, if we use this footer everywhere I'll rename it definately

components/FooterHome.vue
```diff
@@ -0,0 +1,93 @@
...
+            <img src="/img/rogue-amoeba-logo.png" alt="">
+          </a>
+        </div>
+      </div>
+    </div>
+  </footer>
+</template>
+
+<script>
+export default {
```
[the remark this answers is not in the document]
baxgas: it's empty for now, we'll fill it up if there will be some js needed

components/FooterHome.vue
```diff
@@ -0,0 +1,93 @@
...
+
+}
+</script>
+
+<style lang="scss" scoped>
```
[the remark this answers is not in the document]
baxgas: yes, scoped styles apply only to the current component, not to leak out to the general styling (there is globalstyles.scss for general styling)

components/HeadNavigation.vue
```diff
@@ -1,33 +1,34 @@
...
+          About Us
         </NuxtLink>
       </li>
-      <li>Photos</li>
-      <li>Sponsors</li>
       <li>
-        <NuxtLink to="/about/">
-          About Us
+        <NuxtLink to="/shows/">
+          Shows
```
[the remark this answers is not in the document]
baxgas: Good idea. Can be set somewhere yes, but Nuxt/Vue doesn't provide admin interface in itself. Maybe in Arcsi, or wp there could be some endpoint with json objects name, URL etc. ?

components/HeaderBlock.vue
baxgas: it's tailwind css classes for styling, similar to bootstrap (but much nicer and modern)

components/HeaderBlock.vue
baxgas: this is the way the current player works also. but any mp3 data stream should work with it yes. we could have some prop also eg. :arcsi-play="true" and handle all the differences between stream/arcsi with it
I was also thinking to have a unified player, we can discuss it on Saturday.

components/HeaderBlock.vue
baxgas: yes, lahmacun time is important because of the schedule (eg. the Gestalt dude has different time-zone)
a question for Saturday?

components/HeaderBlock.vue
baxgas: hm, good idea for the crosshair. also topic for Saturday?
`getNow` can be a mixin function, whynot?

components/RadioPlayer.vue
baxgas: showAlbumArt is the prop we send in (above you had a comment on that line also) – you are right, it is obsolete if we will always use image.
np.now_playing.song.art was done by gammaw with all the if/else cases of availability of image, default azuracast image etc. with fallbacks to show default
nice ideas with other displays, it is also dependable if we have one player for stream/arcsi if we show the chats there... topic for Saturday

components/RadioPlayer.vue
baxgas: @close is the built in method for modals (very clever), it triggers a function in the Modal component

components/RadioPlayer.vue
baxgas: good idea, but first let's decide if arcsi player is also here

components/RadioPlayer.vue
baxgas: this is not modal, but the volume slider, same as the current player on prod lahma

components/RadioPlayer.vue
baxgas: .prevent is preventing the browser triggering the default click event, same as event.preventDefault
mute is also possible now, it was implemented from the beginning :-)

components/HeaderBlock.vue
baxgas: could be a component on it's own for sure

components/ScheduleHome.vue
baxgas: yes, good idea for further developing the structure and UI. for now it is also a good question how and where to show the other days on home page, I would think some dropdown for selecting other days, or scrolling left-right
the other Schedule view is a nice addition to the rework, definitely worth the effort to develop it, now that technically everything is under our fingertips. topic for Saturday!

components/ScheduleHome.vue
baxgas: no data???

components/ScheduleHome.vue
baxgas: not the nicest solution, but it works, and I was really amazed to see that vue is loading these images on-demand and therefore reducing the bandwidth strain

baxgas: we can discuss this further on Saturday how this hover image show is the nicest, and maybe also to add some other meta info to this popup

components/ScheduleHome.vue
baxgas: question for Csenge, let her work also :-)

components/ScheduleHome.vue
baxgas: it is the new Nullish coalescing operator, really makes things easier for handling unavailable data.
if none is returning it just gives empty string I think, but if you can think some edge cases feel free to tell

components/news/Home.vue
baxgas: format is a part of date-fns library for showing the time in different format https://date-fns.org/v2.22.1/docs/format
new Date is a js Date object

components/news/Home.vue
baxgas: htmlDecoder fixes the errors from wordpress json with some extra characters sent as HTML entity codes. you can check what it does if you turn it off in news titles

components/news/Home.vue
baxgas: $emit emits from children to parent component

components/news/Home.vue
baxgas: watcher is a vue way of checking if something is https://vuejs.org/v2/guide/computed.html#Watchers

components/news/Home.vue
baxgas: wp makes different image sizes on upload, thumbnail, medium and large

layouts/default.vue
baxgas: py-8 = padding on y-axis with 8 units (8*0.25rem)

pages/index.vue
baxgas: layouts are different layouts of components, that can be set in pages -> pages use the default layout by default, but you can change the layout to some custom via this property

pages/index.vue
baxgas: yes, the index starts from zero, that' why we need -1 for the total number (newsLimit)

components/ScheduleHome.vue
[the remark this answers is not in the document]
gammaw: I see. How about commenting it out then?

baxgas:
V-if=false is a commented out block. But true, can be removed all
together,since we can get it back via github history

pages/index.vue
pvj: That's good explanation in layman's terms thanks bax!

layouts/default.vue
pvj: So there's also my and px and mx? Thanks!

components/ScheduleHome.vue
pvj: You use it in `showAirCheck` for logic to verify there is some show on air. That is why I asked how we handle it. Is an empty string ~trueish?

components/ScheduleHome.vue
pvj: Agreed and discussed in person!

components/ScheduleHome.vue
pvj: This was one of my favourites during the demo. Great idea & implementation! 

Vue is cool too they really played their part in this based on what you're writing! 

I would like to have more meta info in there for sure!

components/ScheduleHome.vue
pvj: Yeah maybe there's no feed coming from back-end or something. But it can be revisited later in focus

components/RadioPlayer.vue
pvj: Yeah but not w/ click like that at least I don't remember it. Sorry if it has been there all along!

components/RadioPlayer.vue
pvj: Do you think the volume slider could be part of the player modal?

components/RadioPlayer.vue
pvj: Thanks for the teachment

components/RadioPlayer.vue
pvj: Good studiment, thank you! Yeah I vaguely remember the now playing logic this will be lot cleaner.

Issue #39, opened 2021-04-20
Search for Show on Shows-list page


baxgas:
start typing and after 2letters filters out the matches
maybe this component? : 
https://github.com/santiblanko/vue-instant

PR #66, opened 2021-06-06
#39 search for show on shows list page

gygabor:
I created a reusable search/filter input. It only filters on the given array/array of objects. Needs a design and prepare to handle fetching data from the backend.


pages/shows/index.vue
```diff
@@ -33,36 +33,21 @@ export default {
...
-        this.arcsiShows = [newValue]
-      }
-    },
     arcsiShowsList () {
       if (this.arcsiShows) {
-        const arcsiShowsList = [...this.arcsiShows]
-        return arcsiShowsList.filter((show) => {
-          return !(show.archive_lahmastore_base_url === 'off-air' || !show.active)
-        }).sort((a, b) => a.name.localeCompare(b.name))
+        return this.arcsiShows.filter(show => (
```
baxgas, line 37: Thanks for simplifying this and putting the State in one variable, nice implementation!

pages/shows/index.vue
```diff
@@ -33,36 +33,21 @@ export default {
...
       }
       return null
     },
     pastShowsList () {
       if (this.arcsiShows) {
-        const arcsiShowsList = [...this.arcsiShows]
-        return arcsiShowsList.filter((show) => {
-          return !show.active
-        }).sort((a, b) => a.name.localeCompare(b.name))
+        return this.arcsiShows.filter(show => !show.active)
```
baxgas, line 45: Nice!

pages/shows/index.vue
```diff
@@ -1,18 +1,17 @@
...
-    </div>
-    <!-- Make a listing component -->
     <AutoCompleteSearch
       v-model='arcsiShows'
       suggestionAttribute='name'
     />
-    <ShowsLister :shows='arcsiShowsList' />
-    <div class="m-4 text-center title">
-      <h1>Past Shows</h1>
+    <div v-if='arcsiShowsList  > 0' >
```
baxgas: Nice, good to have the check, not to break stuff if server is down or similar. Maybe also to add some message with v-else?

pages/shows/index.vue
```diff
@@ -41,7 +41,7 @@ export default {
       return null
     },
     pastShowsList () {
-      if (this.arcsiShows.length > 0) {
+      if (this.arcsiShows) {
```
baxgas, line 44: Maybe use the optional chaining, eg. `this.arcsiShows?.length > 0` to check all states? I would leave the length, since it might also be empty if the API server hangs

gygabor:
> Have a suggestion for improvement: to merge the `onChange` and `onEnter` methods in `app/components/AutoCompleteSearch.vue` so that it automatically updates the list (show blocks with images/title/description) on typing, and don't need to press enter or choose from the list, the listed shows are filtered dynamically on changing the text in the input. We could omit the current list (white background dropdown from input) altogether in this case. What do you think?
> Some throttle or setTimeout could be used to delay the emitting of the text to the list array.

Do you mean that while the user is typing in the input, it has to update the list of shows on the page?

baxgas:
Yes, exactly. The shows list could be filtered on keyup/change, exactly as
the dropdown list is listing the filtered list

baxgas:
@gygabor ÷ I made this function in a new method here https://github.com/mmmnmnm/lahmacun_rework/commit/f045558e80cc6776adad293f98f19e43921006a7
I think it is much UX friendlier to filter the list immediately and dynamically show the filtered results in the shows list.
@pvj @gammaw if you can also comment on this?

gammaw (approved):
We decided that we'll go with the "live" search style, so go!

Notes: also, we decided to extend the search functionality so that it searches in the whole "universe" of the given context (e.g. for shows: in _all_ shows, not only in the _loaded_ ones), We can add these new features on the way. Also, we decided to have a search box for archive, show episodes, etc., we should add them at some point too.

gammaw:
@baxgas we'll need to do some conflict resolution here, but it doesn't seem too complex. I would probably try the `6.0.4` version of `table` in the feature branch (because that's main already) and go with that if it works. Let me know if I need to re-approve the merge. 

image

baxgas:
@gammaw I think this branch is what is still in-progress by @gordon he can
report on progress
Btw, the versioning of yarn packages can generally be the most recent,
sure.

gammaw:
@baxgas I see but then let's close the PR, don't we?

gammaw:
> @gammaw I think this branch is what is still in-progress by @gordon he can
> report on progress
>

@baxgas are you sure you mentioned the right user? :)

> @baxgas I see but then let's close the PR, don't we?

We can let @gygabor close it. If it's work in progress then I would close the PR for now and create a new one (or reopen this one) when needed (otherwise it's an open PR for the reviewers). Until then, we can continue the discussion in the issue: https://github.com/mmmnmnm/lahmacun_rework/issues/39

gygabor:
> > @gammaw I think this branch is what is still in-progress by @gordon he can
> > report on progress
> 
> @baxgas are you sure you mentioned the right user? :)
> 
> > @baxgas I see but then let's close the PR, don't we?
> 
> We can let @gygabor close it. If it's work in progress then I would close the PR for now and create a new one (or reopen this one) when needed (otherwise it's an open PR for the reviewers). Until then, we can continue the discussion in the issue: #39

I'm a bit confused. I don't understand the intention of this feature. Should I merge this branch or close it? 
I would merge it and then continue implementing it on the news page in a new branch.

gammaw:
I agree, I think we can merge it as is now, because it implements a meaningful functionality. We can continue on improving the feature with the extended search (among all shows) in a new branch (or using the same branch but a new PR eventually). This was my comment in the approval message too.

Issue #68, opened 2021-06-18
Schedule page with full week schedule


baxgas:
Weekday -> time. Show the active day first, if show on air put an indicator.
Change weekday on click, change slug

PR #85, opened 2021-07-15
#68 schedule page

baxgas:
Schedule page with the weekdays' switcher and the same logic as on homepage


pvj (approved):
Looks good! Do you plan to create weekly full view later?

baxgas:
@pvj
You mean the table view? It is a bit effort to make it, but can be done probably. Maybe the best would be some plugin or framework
Do you think it is effective view to handle the schows, eg. the Gyorsnaszád the only one at 03h, Jun at 11h etc.
Design?

Also, can you please merge this one to the new folder architecture? Thanks!

Üdv-:g

pvj:
@baxgas 
>You mean the table view? It is a bit effort to make it, but can be done probably. Maybe the best would be some plugin or framework
Yeah I mean the table view. I am just thinking out loud 

>Do you think it is effective view to handle the schows, eg. the Gyorsnaszád the only one at 03h, Jun at 11h etc.
Well I definitely didn't think it all the way through. There are probably tons of edge cases and complications. I am not saying we _must_ have _table view_ and nothing else works I just believe that we may/could offer a more visual display that has weekly range instead of text-based daily range we have currently. 

>Design?
Yes, this was great point from you thank you for this and we asked for designs in our last meeting so hopefully we will have something soon!

Issue #86, opened 2021-07-24
Optimize and refactor Arcsi player History


baxgas:
Save Id only once, and then change the playposition and play time infos under that entry.
**Structure**
ArcsiHistory = {
  'episodeID' : [
     playDate: 'time',
     playPosition: 'position'
  ],
  'episodeID' : [
     playDate: 'time',
     playPosition: 'position'
  ],
  ...
}

**Mutation**
newData = {  
  'currentEpisodeID' : [
     playDate: 'currentTime',
     playPosition: 'currentPosition'
  ]
}
Object.assign(ArcsiHistory, newData)

pvj:
Is each object session/user specific? I will try to ask with an example: if I listen to 16:40 and stop but you listen to 23:58 then when I load again is it 16:40 or 23:58?

baxgas:
Handling is Session specific.
User specific would need login

PR #89, opened 2021-08-01
#86 refucktor arcsi play history

baxgas:
The Arcsi History now stores key=episode.id and playDate and seekValue for optimised store object
Thanks @pvj !

### DIFF
--- a/app/components/ArcsiPlayer.vue
+++ b/app/components/ArcsiPlayer.vue
@@ -90,20 +90,26 @@ export default {
       this.setVolume(parseFloat(this.$store.state.player.arcsiVolume))
     }
     this.findIfArcsiSeek()
+    this.$store.commit('player/currentlyPlayingArcsi', this.episode)
   },
   methods: {
     playArcsi () {
       this.togglePlayback()
+      const playHistory = {
+        episodeID: this.episode.id,
+        value: this.progress
+      }
       this.$store.commit('player/currentlyPlayingArcsi', this.episode)
+      this.$store.commit('player/setArcsiProgressHistory', playHistory)
     },
     stopArcsi () {
       this.stop()
-      const arcsiSeekObject = {
-        episode: this.episode.id,
-        value: '0'
+      const playHistory = {
+        episodeID: this.episode.id,
+        value: 0
       }
       this.currentProgress = '0'
-      this.$store.commit('player/setArcsiProgress', arcsiSeekObject)
+      this.$store.commit('player/setArcsiProgressHistory', playHistory)
     },
     volumeBar (value) {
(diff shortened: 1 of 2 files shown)


gammaw:
@baxgas can you help me to understand what's the benefit of the new `Object.assign` solution compared to the `push` one that we had so far?

baxgas:
> @baxgas can you help me to understand what's the benefit of the new `Object.assign` solution compared to the `push` one that we had so far?

@pvj 
The push method adds new entry for every listening item in the Array, the key(episodeID): value(listening info array) object stores per key - which is the unique id of the episode. So there will be only one entry per episode, which is much more optimised than storing all the listen infos. And other infos could be added also later if needed
I thought you commented about optimising it for this structure, no?
This is it:
ArcsiHistory = {
   'episodeID' : [ 
        playDate: 'time',
        playPosition: 'position'
   ],
   'episodeID' : [
        playDate: 'time',
        playPosition: 'position'
   ],
...
}

gammaw:
@baxgas it's me gammaw not pvj :)
I see thanks, so it's a finite key-value store now instead of an infinite history list.
I won't approve it now cause you and pvj know the matter better plus I'd like to try it first, which I'll do cause I have to deal with the CORS issues anyway.

baxgas:
Oh, sorry for misnaming :-)
CORS can be bypassed locally in Firefox, but if you are working now on fixes for it and want to test feel free.
Let me know when it can be merged, I would use this for the new player component also.

Üdv-:g

gammaw:
@baxgas I tried the current version and I wasn't sure how to test the features. In general, what are the main use cases (user stories) where the "memory" of the player would be triggered?  

What I tried is that I 1) visited an episode page and start listening to arcsi, 2) at some point switched to home (without pausing), and 3) navigated back to the same episode page via the show page and hit play again. I was expecting the that episode starts playing where I left it but instead it started from the beginning. Note: I didn't use the browser's controls (e.g., back button).

baxgas:
@gammaw yes, it saves the state on some interaction (pause, seek), but the browser back is not an interaction, it is just quitting the player. That's why it is better to have it somewhere always visible, as the Stream player.
Can we merge this in the main branch? If this is the only problem with it, I think it can be pushed forward in an other "bug" tracker

baxgas:
@gammaw thank you for pointing to this case, I added the fix now, please check so we can merge

gammaw (approved):
I checked on ananas, now the basic functionality seems to work (remembering the playback position for every episode) so I give it an approve. I didn't check the code completely but it's a long-open PR so I think it's good to move it forward. Thanks @baxgas!

baxgas:
Thanks @gammaw !

Issue #88, opened 2021-07-30
Home design update


gammaw:

baxgas:
@pvj - in accordance to streamline the Arcsi API calls, I think it is the best to do it here with the **Arcsi's latest** block, to have a new json endpoint query, eg. arcsiServerURL/latest which should give back just the latest 9 shows in a list
![image](https://user-images.githubusercontent.com/36408851/127765407-87bb9c08-1d1c-490c-85f8-a6a7844d4e12.png)

baxgas:
@pvj - as discussed, also it would be handy to think about the usage of `arcsiServerURL/shows/all` endpoint, to have another optimized endpoint, just listing the necessary `shows` Infos, without all the `episodes` in it that makes the results object quite huge.
So, in the start-query there would be just the basic `shows` infos: name, id, schedule-date...
And then the `episode` (/item/id) data would be fetched just on-demand, as I already implemented it in the arcsi-episode endpage 
As I saw by the progress of the Swagger implementation these should be not a big problem to implement

gammaw:
Can't we now work with the API we have, abstract away the query itself so that we can replace it later on when we have new endpoints without having to change the frontend logic? The query would be a bit complicated at the moment and the call may take a bit more time to complete but it's not that bad I think.

baxgas:
Can be done, but it is a bit overkill to search through the big json file to collect the latest episodes two-three level deep in the hierarchy

PR #90, opened 2021-08-17
#88 home redesign

baxgas:
I made the frontend according to Csenge's designs.
Questions (also asked from Csenge in Figma comments):
- player has no image now, I think it would be good to show it somehow (hover?)
- player play button should turn to pause 
- arcsi episode tags are now dummies
- I wouldn't put all the timezones in Header, makes no sense (the CET is just there for our Schedule timing)


gammaw:
@baxgas I tried to build it (locally and on ananas too) and it gave this error. Any idea? 

image

baxgas:
Did you install the new packages?
yarn install

gammaw:
> Did you install the new packages?
> yarn install

Yes. On ananas I used docker build (see https://github.com/mmmnmnm/lahmacun_rework/blob/main/app/Dockerfile).

baxgas:
Thanks @gammaw for the build check and notion for fixing the error. 
Please let me know if we can merge this state to the main branch

gammaw:
So good to see that the new design is getting shape on the actual draft page! 

Nice structuring with the separate commits too! I skimmed them through quickly and they looked good to me in general. I guess we're in no hurry to merge it into main before our next meeting where we'll discuss some details of home, no? We'll probably need to change things here and there and it's nice to merge a stable version.  

One thing I noticed, under Arcsi's Latest we should add a "latency" similar to what we have for the current website on the shows subpages (so that we don't display an episode before it's premiered).

baxgas:
@gammaw Thank you for pointing out to Arcsi Latest. Besides, I implemented a check for filtering out the not arcsived episodes. 
Also some other fixes and tweaks here-there

pvj (approved):
Let's see it in action!

baxgas:
Thanks!
👍🏻🚀🧞‍♂️

Issue #92, opened 2021-08-28
Make pagination and search for News page


baxgas:
API reference is here:
https://developer.wordpress.org/rest-api/reference/posts/
The query could contain arguments `page`, `per_page`, `search` which could be used for dynamic filtering

PR #94, opened 2021-09-18
#92 news page

baxgas:
Added pagination and Tag filter logic for the News page. Still needs refining, but the basic functionality is working


gammaw (changes requested):
We discussed in the new webpage meeting that we'll have the general search box here too. @baxgas please make the necessary change. Thanks!

baxgas:
@gammaw I removed the Tags selector and finalised the pagination logic (sans any styling). if it is ok now, it could go in master for gygabor to work on it for the search box

Issue #95, opened 2021-09-27
Make fixes for mobile


baxgas:
For now without Mobile version Designs, or final components...

PR #96, opened 2021-09-28
#95 mobile fixes

baxgas:
Mobile fixes with the designs we have now...


gammaw (approved):
@baxgas that's hell of a job! And it looks so pretty on mobile (tried through desktop simulation). Thanks!

Issue #97, opened 2021-09-29
Show page


baxgas:
List all infos from the arcsi endpoint (air-time, frequency)

PR #98, opened 2021-09-30
#97 add Show infos

baxgas:
just added the Show infos from API


gammaw (approved):
Nice! That's what @pvj brought up on Wedn. It seems to be a good first shot to me. I especially like the _Last episode_ section (with the computed date info).

Issue #102, opened 2021-10-04
Show error message if arcsi is Down


baxgas:
Now the whole site breaks...

Check the `$fetchState.error`
https://nuxtjs.org/docs/components-glossary/fetch/#nuxt--212

PR #105, opened 2021-10-07
#102 error pages

baxgas:
Added error messages and a page. Could not really test it locally, but it should work as documented

### DIFF
--- a/app/components/BottomArcsiPlayer.vue
+++ b/app/components/BottomArcsiPlayer.vue
@@ -31,6 +31,10 @@
 import { mediaServerURL } from '~/constants'
 
 export default {
+  validate ({ params, store }) {
+    // Check if arcsiShows exists
+    return store.state.arcsiShows.length
+  },
   data () {
     return {
     }
@@ -68,6 +72,7 @@ export default {
       this.$store.commit('player/isArcsiVisible', state)
     }
   }
+
 }
 </script>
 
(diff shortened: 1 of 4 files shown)


app/components/BottomArcsiPlayer.vue
```diff
@@ -31,6 +31,10 @@
 import { mediaServerURL } from '~/constants'
 
 export default {
+  validate ({ params, store }) {
+    // Check if arcsiShows exists
```
gammaw, line 35: Where will this `validate()` method be called?

pvj (approved):
There really are more than 1 things on this branch which would be easier to understand with details, examples, explanations but I guess it makes sense to cut back on this if there's not much time and you'd rather use that time for development.

Since I can't do much besides reading, I checked the code and tried to understand what you've worked on. Am I seeing correctly that this is part of a work in progress?

From my reading I think I get that:
 - You created error messages in config, 
 - error layouts, error pages, error css,
 - and a validate() method for the arcsiPlayer

So is the plan later to connect player UI with error messages based on the validation? 
And from the config these will be displayed for player then. 
While the error layouts and page and css is for the app level display?

I think if you could give heads-up that this is "part 1" that would be enough for us to see this is in the right context. 

It doesn't seem to break anything so from my side I'm okay w/ merging given that some point later -- when you might not have as much on your hands or if someone can help -- we will get back to finish this.

>In general, I think it's better that the developer always test a new feature before sending a PR.

This is good practice, I agree.

app/components/BottomArcsiPlayer.vue
baxgas: https://nuxtjs.org/docs/components-glossary/validate/#the-validate-method

Issue #103, opened 2021-10-04
Show controls for Arcsi player on mobile lock screen


baxgas:
@gammaw – I think we had some solution for this on the stream player?

gammaw:
> @gammaw – I think we had some solution for this on the stream player?

Yeah, we had this https://github.com/mmmnmnm/lahmacun/issues/182 lengthy issue on the topic.

baxgas:
@gammaw that is exactly how I made it, with MediaSession
https://github.com/mmmnmnm/lahmacun_rework/commit/f7eb8f44d5a37fd6a97bef052b9a93f74c7bca68

gammaw:
@baxgas sure we wanna close it? We had those issues on iPhone, just checked on our test server and it still doesn't seem to wokr properly, e.g., when I switch from arcsi player to the radio player the lock screen still shows the title of the show in the arcsi player and the play/paused statuses didn't seem to correctly update.

PR #111, opened 2021-10-10
#103 mobile lock screen meta tags

baxgas:
I made a fix for the MediaMetadata setter needed to have the currently playing meta data on mobile lock screen. Stream player adds metadata, then queries the API for stream show data changes. Arcsi player adds metadata on play
It should work now, but couldn't check it live because of CORS on my mobile/localhost


gammaw:
@baxgas can you checkout this branch on the test server and try if it works before we merged it?

Issue #99, opened 2021-09-30
Community page mockup


baxgas:
Use wp pages (make if needed)
- Favourite radio sites - page, REST API
- Radio projects – image+title, links to facebook events
- Donate
- ...

PR #113, opened 2021-10-11
#99 communitypage

baxgas:
Content from wp, where we have sections

### DIFF
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -297,6 +297,9 @@ export default {
   },
   watch: {
     volume (volume) {
+      if (isNaN(volume)) {
+        volume = this.$store.player.streamVolume || 50
+      }
       this.audio.volume = Math.min((Math.exp(volume / 100) - 1) / (Math.E - 1), 1)
       this.$store.commit('player/setStreamVolume', volume)
     },
--- a/app/constants/index.js
+++ b/app/constants/index.js
@@ -18,3 +18,4 @@ export const newsURL = 'https://www.lahmacun.hu/wp-json/wp/v2/posts?categories=5
 
 export const aboutUsURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/228'
 export const donateURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/2'
+export const favouriteRadiosURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/2875'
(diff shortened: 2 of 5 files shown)


app/constants/index.js
```diff
@@ -17,3 +17,5 @@ export const newsBaseURL = 'https://www.lahmacun.hu/wp-json/wp/v2/posts?categori
 export const newsURL = 'https://www.lahmacun.hu/wp-json/wp/v2/posts?categories=5&per_page=12'
 
 export const aboutUsURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/228'
+export const donateURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/2'
+export const favouriteRadiosURL = 'https://www.lahmacun.hu/wp-json/wp/v2/pages/2875'
```
gammaw, line 21: @baxgas I guess it would be good to centrally configure the WP site so that we need to change it only once when we redirect the domain.

app/pages/community.vue
```diff
@@ -1,15 +1,153 @@
...
+      <div v-if="donateContent">
+        <h2>{{ donateContent.title.rendered }}</h2>
+        <div class="donate-content" v-html="donateContentResults" />
+      </div>
+    </article>
+    <article id="projects-page" ref="projects">
+      <h2>Projects</h2>
+      <div class="projects-content">
+        Our projects, comming soon...
+      </div>
```
gammaw, line 38: Comming -> coming

gammaw (approved):
Great job, again, @baxgas! Thanks! I added some comments inline and below. Please consider them, but they shouldn't block the PR. 

General comments: 

- Consider adding background for donate lemon (see screenshot), like we have now. 
- I wouldn't include those additional links for the fave radios (see other screenshot).

image

image

app/constants/index.js
baxgas: @gammaw – as for the above, all the content comes from our WP site, if it is changed there it will update here also
nice catch about the endpoints, I made them modular

baxgas:
@gammaw this is a heavy draft page, we will probably have a quite different community page at the end

app/store/player.js
```diff
@@ -1,5 +1,6 @@
 export const state = () => ({
   isStreamPlaying: false,
+  streamVolume: 55,
```
baxgas, line 3: @gammaw this was the missing init

app/components/RadioPlayer.vue
```diff
@@ -297,6 +297,9 @@ export default {
   },
   watch: {
     volume (volume) {
+      if (isNaN(volume)) {
```
baxgas, line 300: @gammaw added a check here also, if some stream break happens

gammaw (approved):
Thanks @baxgas! I haven't tried it again, but it looks good.

PR #114, opened 2021-10-11
#101 add search to news page

gygabor:


gammaw:
@gygabor thanks for the info, makes sense!

> And I was thinking and had an idea what if we create a reusable component from the whole page with search, pagination, design, etc. 

I totally agree!

> I think my PR could be a good starting point. Please check it.

I checked the code, looks good to me. 

Have fun on the sunny side and take care ;)

app/pages/news/index.vue
```diff
@@ -1,30 +1,33 @@
...
   <div class="container">
-    <input
-      v-model="search"
-      class="input"
-      type="search"
-      @input="onChange"
-      :placeholder="placeholder"
-    >
     <header class="flex flex-row items-center justify-between">
       <h1 class="mb-8">
```
baxgas, line 4: Acording to Csenge's design we changed the title block with black background-white letters
```
    <h3 class="title-block">
      News
    </h3>
```
see other main pages also. Maybe this got lost at the merge?

app/pages/news/index.vue
```diff
@@ -65,19 +69,23 @@ export default {
...
+              ? `&per_page=${this.fetchCount}`
+              : ''
+          }${
+            this.search.length > 2
+              ? `&search=${this.search}`
+              : ''
+          }`)
+        this.isLoading = false
+        return callback(response)
+      } catch (e) {
```
baxgas, line 86: We implemented the error handling like this
```
.catch((error) => {
        error({ statusCode: 500, message: 'Stream not available' })
```

baxgas (commented):
Thanks @gygabor ! Works great!
2 small enchancements suggested, if you can put them in that would be great!

app/pages/news/index.vue
```diff
@@ -57,54 +56,70 @@ export default {
...
     }
   },
   async mounted () {
-    this.newsFilteredList = await this.$axios.get(`${newsBaseURL}&per_page=${this.numberOfItems}&offset=0`)
-      .then(res => res.data)
-    this.numberOfTotal = await this.$axios.get(`${newsBaseURL}&per_page=${this.numberOfItems}`)
-      .then(res => parseInt(res.headers['x-wp-total']))
+    this.newsFilteredList = await this.useFetch()
+    this.numberOfTotal = await this.useFetch({ type: 'totalNumber' })
   },
```
baxgas, line 66: Since we are struggling with memory leak issues, maybe the safest is to nuke out the results from memory, sometinkg like this
```
  beforeDestroy () {
    this.newsFilteredList = null
  },
```

baxgas (approved):
Thanks, it can go in

Issue #79, opened 2021-06-19
Photos / gallery 


pvj:
We would like to have dedicated photo / gallery capability

- layout
- page
- appearance on the navbar

baxgas:
It needs some rework from the current solution on the wp page, since the plugin doesn't expose a REST API endpoint.
But, first we should have some form of visual design from Csenge/Marci?

baxgas:
See https://github.com/mmmnmnm/lahmacun/issues/440 after REST is working, connect to nuxt and loop over

PR #120, opened 2021-10-18
#79 gallery rest

baxgas:
Basic Gallery listing and Photos page with Modal. Needs fixing from backend and Design (oc)


app/components/gallery/PictureItem.vue
```diff
@@ -0,0 +1,55 @@
...
+
+<script>
+export default {
+  props: {
+    photo: {
+      type: Object,
+      required: true
+    }
+  },
+  data () {
```
pvj, line 21: is `data` here a required function name if not it might be renamed to something reflecting the return value

app/components/gallery/ListItem.vue
```diff
@@ -0,0 +1,50 @@
...
+export default {
+  props: {
+    gallery: {
+      type: Object,
+      required: true
+    }
+  },
+  computed: {
+    previewImage () {
+    /* TODO -- check if featured image, if not fallback to first image */
```
pvj, line 22: I think it would be perfectly fine like this if we have a fallback default img.

But of course in general we try to add feature images anyway

app/pages/photos/index.vue
```diff
@@ -0,0 +1,40 @@
...
+      </div>
+    </div>
+    <div v-if="$fetchState.error" class="py-32 text-center">
+      Error happened
+    </div>
+  </div>
+</template>
+
+<script>
+import { lahmaGaleriesURL } from '~/constants'
```
pvj, line 20: There's a typo here in galleries

app/components/gallery/PictureItem.vue
```diff
@@ -0,0 +1,55 @@
...
+    <Modal :media="photo.full_image_url" :title="photo.title" :visibility="shadowbox" />
+  </a>
+</template>
+
+<script>
+export default {
+  props: {
+    photo: {
+      type: Object,
+      required: true
```
pvj, line 18: Could we define more exact props here that we use in the template? 

So is vue props here something like the parameters of an object during initialisation? In that case I think it would be clearest to here define `title` `full_image_url` etc. that are referenced in the template. Or is this already defined somewhere that I missed? Or did I go the wrong way? Curious to hear your thoughts! Thanks.

pvj (approved):
Some non-blocking comments and questions thanks for the hard & consistent work!

Issue #131, opened 2021-10-29
Add pause button, fix arcsi pause state trigger


baxgas:
+++ rework images to assets folder for better compiling

PR #132, opened 2021-10-30
#131 improvements

baxgas:


gammaw:
@baxgas I checked the changes (in code and tried locally), they look good to me. Any reason why it's a draft PR yet?

baxgas:
@gammaw yes, I want to make a small function grouping change, so it is more modules, and the events can call them separately

baxgas:
@gammaw I reworked the scripts a bit, so it is more modular and the audio play/pause events just trigger the necessary setters. Added a bit bottom margin to photos lists, since they were a bit sticking to bottom

gammaw (approved):
Nice updates, I tried to test the changes and checked them code-wise too. Thanks!

Issue #91, opened 2021-08-19
Sandclock progress animation


baxgas:
Asked from Csenge static frames to animate the progress.

PR #133, opened 2021-11-01
#91 sandclock anim

baxgas:
Made the frame-by-frame anim in 10 stages


gammaw (approved):
Nice start anim! Thanks!

Issue #121, opened 2021-10-22
Photos gallery list - show featured image if exists, add previous/next image for shadowbox


baxgas:
add a method to check/fetch featured image if exists. If not, use first image in gallery (like current method)

👉 So is vue props here something like the parameters of an object during initialisation? In that case I think it would be clearest to here define title full_image_url etc. that are referenced in the template. Or is this already defined somewhere that I missed? Or did I go the wrong way? Curious to hear your thoughts! Thanks.

+++ add previous/next image navigation for shadowbox

PR #137, opened 2021-11-01
#121 galery improvements

baxgas:


app/components/RadioPlayer.vue
```diff
@@ -113,7 +113,7 @@
         </div>
 
         <div class="sand-clock">
-          <IconSandclock :progress="time_percent" :live="np.live.is_live" />
+          <IconSandclock :progress="time_percent" :live="np.live.is_live.length ? true : false" />
```
gammaw, line 116: @baxgas why is this?

gammaw (approved):
Thanks for this too! I tested locally and scanned through the code changes. I discovered some improvements, e.g., the arrow navigation within the modal carousel view, very nice!, but it would help if you could highlight the changes in the description of the PR (like you often do), especially if the changes affect multiple files and add a lot of code. 

Also, I guess we should start documenting components for what they are doing, who is using them etc. Do you know of any best practice for that in these type of applications, e.g., a comment in the first line of the component file, in a given format? For example, `ListeItem`, `PictureItem`, `ModalGallery`, if you're not the implementor, then it may not be trivial to see what the hierarchy is. 

A specific question: why is `ModalGallery` not under `components/gallery`?

baxgas:
Nice catch! Since it is reworked, gallery specific modal it could be under
gallery/modal and GalleryModal component name
The sandclock i found that it didn' catch the is_live state, sofix is for
that

baxgas:
@gammaw please review again, I implemented the new functions and there was also a merge conflict with master that needed resolving

gammaw (approved):
Seems all good. Thanks for the structure update and the descr -> caption fix!

Issue #134, opened 2021-11-01
Implement show type on the endpage with icons


baxgas:
`language: "music", language: "hu_hu",`
Check/fix also on Show level in Arcsi admin

PR #140, opened 2021-11-08
#134 add icons to languages

baxgas:


gammaw (approved):
Good initial addition! We may polish it design-wise, the listing in the metadata block looks very basic now.  

image

PR #142, opened 2021-11-11
#138 Csenge new comments

baxgas:
Just some small fixes after Csenge's comments on new version


app/components/RadioPlayer.vue
```diff
@@ -112,7 +112,7 @@
           </div>
         </div>
 
-        <div class="sand-clock">
+        <div v-if="false" class="sand-clock">
           <IconSandclock :progress="time_percent" :live="np.live.is_live" />
```
gammaw, line 116: Do we disable the sand clock completely?

gammaw (commented):
@baxgas see inline comment about the sand clock.

app/components/RadioPlayer.vue
baxgas: Csenge told me also she finds it unneeded and distracting

baxgas: Otherwise it is of course open to discussion on the next meeting

gammaw: Oh, that's a bold change :) I would say, let's wait until the meeting before we merge this change. And if we decide to get rid of the sand clock, I think it's better that we delete it from code (we have it in the history in case we want to bring it back some time in the future).

pvj (commented):
Let's keep the hourglass!!!!

For everything else I'm in favor of merging this.

pvj (approved):

gammaw (approved):
Save hourglass succeeded! :)

Issue #143, opened 2021-11-14
Fix Errors reported by Sentry 


baxgas:
**2. Message**
Error: AbortError: The play() request was interrupted by a call to pause().

**1. Message** 
t is not a function
/_nuxt/85e59aa.js in checkNowPlaying/< at line 1:96113
.catch((function(t){t({statusCode:500,message:"Stream not available"})}))
axios error handling?
https://code.luasoftware.com/tutorials/vuejs/nuxt-fetch-handle-error/

PR #144, opened 2021-11-14
#143 sentry fix

baxgas:
Fixes the catch block handler to properly send error object to nuxt

### DIFF
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -447,7 +447,8 @@ export default {
         }
         // Vue.prototype.$eventHub.$emit('np_updated', npNew);
       }).catch((error) => {
-        error({ statusCode: 500, message: 'Stream not available' })
+        console.log(error)
+        this.$nuxt.error({ statusCode: 500, message: 'Stream not available' })
       }).then(() => {
         clearTimeout(this.np_timeout)
         this.np_timeout = setTimeout(this.checkNowPlaying, 15000)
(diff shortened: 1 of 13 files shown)


pvj (approved):
Looks good! If we release this after sentry integration we can  check live whether it's fixed.

One additional request, I don't know if these are ignored in prod but if not I don't want to spam the console with errors, so if you verified the logs can you remove console logging lines? 

Thanks for the good work figuring out & solving this! Gg

gammaw:
@baxgas general question: how can you map the location of the error across JS and Vue? In Sentry I only see a rough location in the generated JS (e.g., inside the funtion `checkNowPlaying` and the generic symbol name `t`) but I couldn't figure our at which specific occurance. 

```
TypeError: t is not a function
  at checkNowPlaying/< (/_nuxt/85e59aa.js:1:96113)
```

baxgas:
@gammaw  Sure, it was the catch error function which didn't point tot he global $nuxt context object 
image
Fixed it in this PR.
Yes, the prod script is minified, so the inline code is a bit difficult to decipher, but it was quite obvious
`.catch((function(t){t({statusCode:500,message:"Stream not available"})}))`

gammaw:
And how do you know that it's `t`'s occurence within the catch block and not somewhere else? 

I see this in Sentry and `t` occurs multiple times. 

image

Like pvj writes, we'll see if the error disappears, I'm just curious how you knew that the problem is in the error case.

Issue #104, opened 2021-10-04
App down on out of memory error 


gammaw:
Here's the error log for crash. 

Further circumstances: 
- This time it happened within 24h after a fresh launch. 
- The container also quits after this error. 

![image](https://user-images.githubusercontent.com/28865986/135873723-4fb4ec9e-55ca-4914-aa79-3de46bda1b0e.png)

FYI @pvj @baxgas @gygabor I didn't know of another issue for this so I created this one.

pvj:
I've added an `--inspect` option too to the `start` script in on our staging env and re-built and re-started the container. This way it's possible to get profiles from the browser dev console. Will try to do that and locate the leak. 

Thanks for the monitoring @gammaw !

baxgas:
@pvj @gammaw 
See this please
![image](https://user-images.githubusercontent.com/36408851/136751490-03715903-0d90-4293-ae20-b7bbf0a10e2a.png)
https://jcutrer.com/javascript/yarn-build-exit-code-137

gammaw:
Our current experiments - highlights:

- Links prefetch disable 
- Webpack watcher ignore 
- max old space increase
- memwatch

gammaw:
@pvj thanks for these tool additions! For memwatch, I was wondering: 

- As far as I understand, we can now visually see the memwatch reports when running `docker-compose up`  (without the `-d` option), which only applies for the user who runs the command, but other team members cannot not have access to the reports. More importantly, we lose the reports if the app shuts down (either because we shut it down or out of mem). Is there an easy way to write these reports on disk (like a file log)?  
- Again, as far as I understand, all `memwatch` options we use are passive "monitors" in nature, except `gcAfterEvery` which changes the default behavior of gc. At this point, we're trying to diagnose what is happening and this setting may affect the native behavior, so wouldn't it be better to disable it?

gammaw:
We discussed today that 1) @baxgas will create a one-night log with memwatch entries together with the proxy traffic (by running `docker-compose up` on test) and 2) @baxgas will try three stripped-down versions (on respective branches) so that there is i) no arcsi connection, ii) no WP connection, iii) no Azuracast connection. At the end of each experiment we should record the mem status too.

gammaw:
[...]

PR #145, opened 2021-11-15
Performance and run-time error monitor integration for #104

pvj:
What happened?
Trying to guard against and catch the memleak we added monitoring solution for Sentry. It's mainly just config and deps and env files & changes. 

What did I do?
Added the needed nuxt libraries and nuxt config

What's next? 
Will try to add perf metrics, apdex etc.


baxgas (approved):

pvj:
@gammaw I was (somewhy) under the impression performance tracing would enable us just that but it seems like it is not the case

Issue #149, opened 2021-11-17
Schedule blocks more informative


baxgas:
![image](https://user-images.githubusercontent.com/36408851/142217105-5a1b519e-fdcf-4798-b3e2-38b519076760.png)

PR #151, opened 2021-11-18
#149 schedule block

baxgas:
Big rework of Schedule Show blocks, showing big OnAir episode, and the last Arcsi episode for the others

### DIFF
--- a/app/components/schedule/Fullitem.vue
+++ b/app/components/schedule/Fullitem.vue
@@ -1,0 +1,137 @@
+<template>
+  <div :class="showAirCheck(0, show.name) ? 'dayblock onair' : 'dayblock'">
+    <div class="onairshow">
+      <span class="text-red-600">●</span>
+      On Air
+    </div>
+    <div class="show-basic-infos">
+      {{ removeSeconds(show.start) }}
+      <img src="@/assets/img/arrow-schedule.svg" alt="" class="inline-block w-10">
+      {{ removeSeconds(show.end) }}
+      <span class="hidden md:inline-block"> - </span>
+      <nuxt-link :to="'/shows/' + show.archive_lahmastore_base_url" class="block md:inline-block">
+        <b>{{ show.name }}</b>
+      </nuxt-link>
+      {{ showAirCheck(0, show.name) && ' | ' + streamEpisodeTitle }}
+    </div>
+  </div>
+</template>
+
+<script>
+export default {
+  props: {
+    show: {
+      type: Object,
+      required: true
+    }
+  },
+  data () {
+    return {
+      interval: null,
+      nowPlaying: {}
+    }
+  },
+  computed: {
+    isTouchEnabled () {
+      return ('ontouchstart' in window) || (navigator.maxTouchPoints > 0) || (navigator.msMaxTouchPoints > 0)
+    },
+    streamShowTitle () {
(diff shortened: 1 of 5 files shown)


app/components/schedule/Fullitem.vue
```diff
@@ -0,0 +1,223 @@
...
+    streamShowTitle () {
+      if (!this.nowPlaying.now_playing) {
+        return false
+      } else if (this.nowPlaying?.live?.is_live) {
+        return this.nowPlaying?.live?.streamer_name
+      } else {
+        return this.nowPlaying?.now_playing?.song.artist
+      }
+    },
+    streamEpisodeTitle () {
```
pvj, line 97: These are some wacky things!! 

If we use these in other places should we create a global helper component where these can be written and then imported anywhere?

app/pages/schedule.vue
```diff
@@ -4,39 +4,23 @@
...
-                <div v-if="!isTouchEnabled" class="show-image" :style="{ backgroundImage: `url(${show.cover_image_url})` }" />
-              </div>
+      <div class="mb-4 border-b days">
+        <ul>
+          <li v-for="(day, dayIndex) in dayNames" :key="dayIndex">
+            <div class="px-4 py-2" :class="selectedDay === dayIndex && 'bg-white'" @click="changeDay(dayIndex)">
+              <h4 class="block">
+                {{ day }}
+              </h4>
+              {{ $moment(todayDate).add(dayIndex, 'days').format('MMM Do') }}
```
pvj, line 14: Hopeful for `$moment` helping where `date-fs` failed us!!!!

app/pages/schedule.vue
```diff
@@ -153,6 +103,13 @@ export default {
       this.$refs[dayindex][0].classList.add('block')
       this.$refs[dayindex][0].classList.remove('hidden')
       this.selectedDay = dayindex
+    },
+    checkNowPlaying () {
+      this.$axios.get(this.streamServer).then((response) => {
+        this.nowPlaying = response.data
+      }).catch((error) => {
+        console.log(error)
```
pvj, line 111: Maybe something else instead of console log? For example we can create custom errors for Sentry like this:
 - in vue components: `this.$sentry.captureException(new Error('example'))`
 - in async: `catch (error) {
    $sentry.captureException(error) }`
 - in middleware: `process.$sentry`
 
 See more here: https://sentry.nuxtjs.org/guide/usage

pvj (approved):
Very good job Gas! Unbelievable I'd say, can't keep up with the PR's 

I made two comments. I don't want to force these changes on you but I believe in one case `console.log` one we could switch & improve w/ passing it to Sentry instead

app/components/schedule/Fullitem.vue
baxgas: Yes, it would be possible to have a global mixin, but we use this at two places in my knowledge, and it is quite stream centered

app/pages/schedule.vue
baxgas: much simpler also!

app/pages/schedule.vue
baxgas: wow, yess, we can, definitely! but I think it is better to do it globally, for all the error messages, I'll open a new issue with this

PR #153, opened 2021-11-19
More measurements for app performance

pvj:
Trying to locate root of #104 
- Added sentry performance tracing 
- Enabled nuxt-memwatch module (see all of its options here: https://github.com/pimlie/nuxt-memwatch#module-options)


baxgas (approved):
Thanks @pvj !

Issue #154, opened 2021-11-19
All error messages go to Sentry


baxgas:
Maybe something else instead of console log? For example we can create custom errors for Sentry like this:
- in vue components: this.$sentry.captureException(new Error('example'))
- in async: catch (error) { $sentry.captureException(error) }
- in middleware: process.$sentry

See more here: https://sentry.nuxtjs.org/guide/usage

PR #160, opened 2021-11-21
#154 connect sentry to catch, adjust error page

baxgas:


app/components/schedule/Home.vue
```diff
@@ -95,8 +95,9 @@ export default {
       this.$axios.get(this.streamServer).then((response) => {
         this.nowPlaying = response.data
       }).catch((error) => {
-        console.log(error)
-        this.$nuxt.error({ statusCode: 500, message: 'Schedule not available' })
+        console.log('Network interrupted stream. Automatically reconnecting shortly...', error)
+        this.$sentry.captureException(new Error('Stream interrupted ', error))
+        this.interval = setTimeout(this.checkNowPlaying, 15000)
```
molnar-a, line 100: If I understand this correctly, if the nowPlayingData request fails, we check again only once after 15 seconds. Is this the expected functionality? Should we reset the interval timer if the 15 second check results successfully?

baxgas: @molnar-a Thanks, I made a clear on succesfull response

app/components/HeaderBare.vue
```diff
@@ -0,0 +1,31 @@
+<template>
```
pvj, line 1: Nice to have a Bare bones template. Good idea!

Maybe we can use this to generate a non-js page for those hackers out there blocking scripts on the web in the future?

app/components/RadioPlayer.vue
```diff
@@ -453,7 +453,8 @@ export default {
         }
         // Vue.prototype.$eventHub.$emit('np_updated', npNew);
       }).catch((error) => {
-        console.log('Network interrupted stream. Automatically reconnecting shortly...')
+        console.log('Network interrupted stream. Automatically reconnecting shortly...', error)
```
pvj, line 456: If we log everything to sentry do we have reason to keep these console.logs?

Is the point not to remove all of them and replace them with sentry logging?

app/components/schedule/Home.vue
```diff
@@ -44,6 +44,7 @@ export default {
       showsByDate: [],
       dayNames: ['Monday', 'Tuesday', 'Wednesday', 'Thursday', 'Friday', 'Saturday', 'Sunday'],
       interval: null,
```
pvj, line 46: Can we rename this to something else that tells us that this is about the clock 1 minute intervals? clockInterval? clockMinute? I trust you w/ naming it anything but after introducing `recheckInterval` this original `interval` confused me a bit

app/pages/index.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <div>
     <section class="grid-cols-2 mb-16 md:grid home-top">
-      <div class="bg-white">
+      <div v-if="newsListState" class="bg-white">
         <NewsHome :news="sortNews[newsStart]" @changenews="changeIt($event)" />
       </div>
-      <div>
+      <div v-if="sortShowsForSchedule">
```
pvj, line 7: For me it would be simpler to directly see what the conditional is doing instead of putting the method name here that we use below as well, does this go against some best practices? What do you think?

Can we just write this conditional rendering directly `v-if="this.arcsishows"` ? Same for `newsList` & `newsListState`

app/pages/news/tags/_tag.vue
```diff
@@ -11,38 +11,40 @@
...
+        this.$sentry.captureException(new Error('Tags not found ', error))
+        this.$nuxt.error({ statusCode: 500, message: 'Tags not found' })
       })
     this.tagsPosts = await this.$axios.get(`${contentApiURL}/posts?tags=${tagId}&per_page=100`)
       .then(res => res.data)
       .catch((error) => {
         console.log(error)
-        this.$nuxt.error({statusCode: 500, message: 'Tags not found'})
+        this.$sentry.captureException(new Error('Tags posts not found ', error))
+        this.$nuxt.error({ statusCode: 500, message: 'Tags not found' })
```
pvj, line 44: Technically saying not found is correct error but maybe to make it simple for users we could just return `No tags` message or nothing at all, null. 

I think this could be also changed to `statusCode: 404` (or the above error on `L37`?)

app/pages/news/tags/index.vue
```diff
@@ -26,32 +26,34 @@
...
     }
   },
-  async fetch() {
+  async fetch () {
     this.recentNews = await this.$axios.get(`${newsBaseURL}&per_page=100`)
       .then(res => res.data)
       .catch((error) => {
         console.log(error)
-        this.$nuxt.error({statusCode: 500, message: 'News not found'})
+        this.$sentry.captureException(new Error('New not found ', error))
```
pvj, line 43: small typo here!

app/pages/news/tags/index.vue
```diff
@@ -26,32 +26,34 @@
...
+        this.$nuxt.error({ statusCode: 500, message: 'News not found' })
       })
     if (this.allTags) {
       this.allTagsList = await this.$axios.get(`${tagsURL}?include=${this.allTags.toString()}&per_page=100`)
         .then(res => res.data)
         .catch((error) => {
           console.log(error)
-          this.$nuxt.error({statusCode: 500, message: 'Tags not found'})
+          this.$sentry.captureException(new Error('Tags not found ', error))
+          this.$nuxt.error({ statusCode: 500, message: 'Tags not found' })
```
pvj, line 52: Do we really need this twice? Should this be `500` or `404`?

pvj (dismissed):
Enormous change request touching almost all parts of the app. 

Huge work, thanks!!! 

I had some questions and one big request -- removing `console.logs()`. Maybe I am missing something but I don't see reasons to keep them and let them print our stuff to any browser.

Since you made all these heavy changes I tried to pay as much attention to the code as possible, and commented after thorough consideration. 

So I understand that all these comments at once will be huge rainfall on you. Please don't read & answer  & tackle them all at once, I hope by taking it one-by-one it won't be so burdensome and daunting to discuss the PR. 

Gogo

app/components/HeaderBare.vue
baxgas: it is just the header that is shown on error page, without the menu etc.

app/components/RadioPlayer.vue
baxgas: I don't know, I think it is good for debugging, but to end-users it is not relevant at all

app/components/schedule/Home.vue
baxgas: warum nicht...

app/pages/index.vue
baxgas: it is needed to check if this computed object exists at all, before trying to render. otherwise it throws an error if it is still not calculated before rendering
it is a general practice in vue/nuxt, yes

app/pages/news/tags/_tag.vue
baxgas: I don't know the best practices for these messages, I'll rely on your opinion

app/pages/news/tags/index.vue
baxgas: Generally, I think 500 is the server error (in this case Wordpress API server), and 404 is the error if the route in the site (in this case nuxt) is not existing, or am I wrong? @pvj

baxgas:
@pvj thank you for the comments, I replied to them, generally I removed all the console.logs, so everything goes to sentry now.

baxgas:
@gammaw @pvj 
I updated this branch, it can go in main

gammaw (approved):
It's an old (and useful) one, and as far as I can see the original change request has been fulfilled. Thanks!

Issue #163, opened 2021-11-22
Add correct sorting for the Schedule


baxgas:
Add sort by play_date before checking

PR #164, opened 2021-11-22
#163 correct the order of show items before checking the latest

baxgas:


pvj (approved):
Looks like straightforward implementation and good improvement to handle this! M-eh-erh-geh

gammaw:
@baxgas I just noticed that we show latest episodes on the schedule page. How come? Arcsi will not always tell us what will be aired in a slot, and even if a fresh episode is already in arcsi, we don't want to show it in advance. Sorry if I'm fundamentally missing something :)

baxgas:
@gammaw
Yes, the latest uploaded which is in best case the currently airing, we
discussed with J to have a flag from arcsi which is the broadcasted
(currently buggy) and it will be from there at the end.
As it gets the episode at end page it will check if it is archived or not,
and show tge player accordingly

gammaw:
@baxgas after doublechecking with @pvj the requirements could be this: 

- Display episode data for ON AIR show (because we can be certain).
- Display episode data for shows _before_ the ON AIR show (because we know it won't change, note that it's only shows on that day).
- Do not display episode data for everything else (because we cannot be sure what will be played, display general show data instead). 

We also said that we could fix the behaviour in this PR, but I'm not sure it's possible (as far as I remember GitHub, will treat a PR merged even you add new commits to it).

Issue #166, opened 2021-11-24
Rework the Slider plugin to custom solution


baxgas:
Also add drag functionality
https://codepen.io/bushblade/pen/ZEpvzbK?editors=0010

PR #167, opened 2021-11-24
#166 slider rework

baxgas:
Removed the component and made a custom solution. 
Sorry, no drag-n-drop here


app/components/arcsi/Latest.vue
```diff
@@ -1,24 +1,44 @@
...
+        class="latest-nav previous"
+        :style="{ top: episodeWidth / 2 + 'px' }"
+        @click.prevent="previousBlock"
+      >
         <img src="@/assets/img/arrow-left.svg" alt="">
-      </div>
-      <div slot="button-next" class="swiper-button-next">
+      </a>
+      <a
+        v-show="sliderPosition < numberOfEpisodes - visibleEpisodes"
```
pvj, line 33: What happens when there is only 1 more next episode? 

`sliderPosition = 8`
`numberOfEpisodes = 9`
`visibleEpisodes = 3`

Then `8 < (9-3)` doesn't hold.

Does that mean we can not reach the last one? 

Maybe I misunderstand which item is in focus, maybe it is never 3rd visible episode but the middle one. Still `7 < (9-3)` doesn't hold. 

I believe I am making some false assumptions on the calculations you are doing so forgive me if it is stupid to ask.

app/nuxt.config.js
```diff
@@ -84,7 +83,7 @@ export default {
     publish_release: true,
     sourceMapStyle: 'hidden-source-map',
     config: {
-      release: process.env.GIT_COMMIT_SHA,
+      release: process.env.GIT_COMMIT_SHA
```
pvj, line 86: Good catch thank you!

pvj (approved):
Hi @baxgas Thanks for the nice work on quickly implementing this from scratch! That's a real feat and I think we will benefit of it as we develop further the app. 

I couldn't find issue or give constructive feedback re: js code, nice work there! 

I tried to do some math on the math equations and I might have been doing the wrong math but I couldn't firmly verify all cases. Can you please check my example in review comments?

app/components/arcsi/Latest.vue
baxgas: @pvj the slider position starts from index 1, so 8 = 9
Nice catch about the last item resizement! I'll make a fix for it

app/nuxt.config.js
baxgas: Just a different linter (PHPStorm) :-)

pvj (approved):
Thanks for clearing it up I always forget that index is `1` in this weird universe :D 

Thanks for the changes on sliderPosition, looks good now on all accounts I think

gammaw:
> Removed the component and made a custom solution.

@baxgas what was the reason for doing so? Was it a vulnerable dependency, or a design issue? Sorry, I must have missed it. 

> Sorry, no drag-n-drop here

What do you mean by drag-n-drop?

baxgas:
@gammaw it was a bit uncompatible with nuxt, working natively with vue, 
Linter complained about it and Ádám also found out there was some compatibility issues with it

PR #169, opened 2021-11-26
#104 mem-leak test for client-only schedule

baxgas:
I think this works fine and solves the mem-leak.
Thanks for the tips for solving it!


pvj (approved):
I'm a duck, I'm a firework, I'm clapping, and bobbing my head.

Issue #170, opened 2021-11-30
Add footer for all pages


baxgas:
Follow Contact Sponsors

PR #171, opened 2021-11-30
#170 footer rework

baxgas:


app/components/footer/Bottom.vue
```diff
@@ -0,0 +1,87 @@
...
+          <h5>Contact</h5>
+          <p>
+            <a href="mailto:contact@lahmacun.hu">contact@lahmacun.hu</a>
+          </p>
+          <h5 class="mt-4">
+            Impressum
+          </h5>
+          <p>
+            <NuxtLink to="/impressum/">
+              Lahmacun impressum
```
gammaw, line 41: I would just have "Impressum" which is a link to the Imrpessum page.

gammaw (approved):
Good to see the footer being added to all pages!

For the WP content of the Impressum page: I realised that the hrefs do not appear. Can we fix it?

Issue #172, opened 2021-11-30
Schedule page rework


baxgas:
On Schedule page show only the shows info (for not currently ON AIR)
Make a copy of current latest episode Schedule view and group days Monday-Sunday (latest Archives) - link on Archives page?

PR #173, opened 2021-11-30
#172 rework schedule and add weekly archives page

baxgas:
@gammaw 
here is the rework of the schedule, the main schedule page is showing now the show infos and onair data,
and there is a new route with a day/latest episodes view at `../weekly-archives`


gammaw (approved):
Big effort! The separation between schedule and weekly archive view makes sense too me as a proposal. We can discuss it with the others. Same goes for the calendar icon on the archive page. Nice shortcut!

Issue #175, opened 2021-12-03
Rework Community page


baxgas:

gammaw:
@baxgas there seems to be a bug on the community page, see code part below. What happens is that if I add a featured image to the labs subpage (on WP), then the community page will render wrongly (texts disappear etc.)

https://github.com/mmmnmnm/lahmacun_rework/blob/31c039accd276d4d5a8a12460da8c14a68867b3c/app/pages/community.vue#L277

PR #176, opened 2021-12-03
#175 rework community

baxgas:


gammaw (approved):
Good PR to get it rolling with the community + donate menus and w/ their initial contents! 

The community page looks very text heavy now but I see that we can add pictures in WP to the respective content pages. Plus, we'll have a design for it I guess.

Issue #180, opened 2021-12-05
Quick fix mobile view of header-player


baxgas:
for below 360px
![image](https://user-images.githubusercontent.com/36408851/144750068-561ee654-f0cb-4e6a-87a2-a4e8478005dd.png)

baxgas:
+ mobil view checks 
![image](https://user-images.githubusercontent.com/36408851/144753132-2b3db0ef-71a3-48ea-b6a5-49b15eefc2ef.png)

PR #181, opened 2021-12-05
#180 mobilefixes

baxgas:


gammaw (approved):
Nice clean-up on the messy mobile home! Since we don't yet have a ready mobile design, it's a valid shortcut to remove the sand clock on small mobiles for now.

Issue #177, opened 2021-12-05
Arcsi player won't start until seeking into track (iPhone only)


gammaw:

baxgas:
@gammaw
Schau mal
https://stackoverflow.com/questions/10985640/html5-audio-callback-fails-on-safari-ios

gammaw:
@baxgas thanks! I encountered the bug on Chrome not on Safari but I'll check your code changes in the PR.

PR #182, opened 2021-12-06
#177 fix safari bugs

baxgas:
@gammaw 
I searched out in the documentation how to hack out the Safari audio play event, if you can check also on iPhone please

### DIFF
--- a/app/components/arcsi/Player.vue
+++ b/app/components/arcsi/Player.vue
@@ -7,6 +7,7 @@
         preload="metadata"
         :title="episode.shows[0].name + ' - ' + episode.name"
         :src="source"
+        :autoplay="isSafari"
         @play="setPlayState()"
         @pause="setPauseState()"
         @loadedmetadata="getDuration()"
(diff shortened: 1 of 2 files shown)


app/components/arcsi/Player.vue
```diff
@@ -169,6 +172,9 @@ export default {
     }
     this.setMetaData()
     this.$store.commit('player/currentlyPlayingArcsi', this.episode)
+    if (!(navigator.vendor.match(/apple/i) || '').length > 0) {
```
gammaw, line 175: @baxgas don't we want to pause if we have an Apple device (so `isSafari `instead of `not isSafari`)? Btw, can't you use iSafari here?

gammaw (commented):
@baxgas nice finding! 

A general comment: for me the key seems to be from [your stack overflow link](https://stackoverflow.com/questions/10985640/html5-audio-callback-fails-on-safari-ios) this: 

> Apple does not allow you to call .play on any audio until user interaction

But then, we could try to detect if the device is Apple instead of checking on Safari, don't we? 

I added a specific comment where you changed the mounted routines.

app/components/arcsi/Player.vue
baxgas: @gammaw  I think on iOs all the browsers use Safari engine, just the "skin" is different, so the check should work for all. If you can try on a real device that would be the best

gammaw: Ok, I try to build the branch on the test server. 

But before doing so: 

You think that `vendor` would be Apple for the Chrome app on iPhone? 

But what I meant by my original question was that don't you want to pause the player on Apple devices? It seems to be reversed logic logic now.

gammaw (approved):
The arcsi player seems to work now in Chrome and Safari on an iPhone X. Cool!
The lock screen play control still not working though.

Issue #183, opened 2021-12-06
Prelaunch tasks


baxgas:
- [x] Rare shows logic implement
- [x] Remove arcsi weekly view and show latest episode link in Schedule
- [x] check footer: text, links, sponsors, collaborators @gammaw 
- [x] @baxgas please remove all sponsors for now
- [x] check Community page content @gammaw ?
- [x] Shows page 4 in a row and more compact (no white back, just title?)

gammaw:
> check footer: text, links, sponsors, collaborators @gammaw

@baxgas it's kind of done I guess. I asked you here https://github.com/mmmnmnm/lahmacun_rework/pull/171 to update the Impressum text on the hyperlink and elsewhere (Trello) that you delete all sponsor logos. Can you do it? 

>  check Community page content @gammaw ?

I need to find 1-2 more pictures but otherwise we're done contentwise. I asked you to modify the template so that the pictures are used from WP. What do you think of it?

baxgas:
@gammaw yes, this is a check list, so we collect all the infos

gammaw:
@baxgas thanks! It's deliberately not a PR yet, right?

baxgas:
@gammaw see the dependencies check-list. 
gygabor's list is a separate task, pvj's is a trello task IMHO. So, you think we should leave them out from here?

gammaw:
Thanks @baxgas for the PR https://github.com/mmmnmnm/lahmacun_rework/pull/184, I'll check it asap.

PR #184, opened 2021-12-08
#183 finalise before lunch

baxgas:


gammaw (approved):
Nice! The community page looks much nicer now. Cool additions on the shows page too. Thanks for the Impressum change, it looks a bit strange compared to the other links in the footer but that's design too. I see the the sponsor logo is still there, can you remove it in a later change?

PR #185, opened 2021-12-11
GA4 script added to head

gammaw:
Please review the addition of the GA4 script for traffic monitoring of the beta site after the Luca launch.

Notes:
- I know of this plugin https://github.com/MatteoGabriele/vue-gtag but it seems to be not very fresh plus it's a dependency (w/ possible vulnerabilities) so I went for the native solution:
- https://nuxtjs.org/docs/concepts/views/#app-template

### DIFF
--- a/app/app.html
+++ b/app/app.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html {{ HTML_ATTRS }}>
+  <head {{ HEAD_ATTRS }}>
+    {{ HEAD }}
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-BBW6L3XL7Z"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+      gtag('config', 'G-BBW6L3XL7Z');
+    </script>
+  </head>
+  <body {{ BODY_ATTRS }}>
+    {{ APP }}
+  </body>
+</html>


pvj (approved):
Technical implementation looks good

I don't know if templates are ok or not for the front-end, I am not involved enough I think to comment on that. Personally, I like it and it never hurts to try these features and see what sticks. 

Thanks!

baxgas (commented):
@gammaw I don't think this is the best way, but if you found the guide at some specific place? 
I would think this is the right place for the head script tags https://github.com/mmmnmnm/lahmacun_rework/blob/31c039accd276d4d5a8a12460da8c14a68867b3c/app/nuxt.config.js#L11
see this https://reactgo.com/nuxt-add-script-to-head/
but I might be wrong... 
good you didn't use the plugins! let's keep things simple and clean as much as possible!

baxgas:
@gammaw I didn't get any triggers for the play events, I think they need also the setup. for GA v3 it was these, don't know if they are valid or  changed since 
https://github.com/mmmnmnm/lahmacun_rework/blob/31c039accd276d4d5a8a12460da8c14a68867b3c/app/components/RadioPlayer.vue#L412
https://github.com/mmmnmnm/lahmacun/blob/98bbc2b199448174cbc4783ce24c5bbabfa94665/wp/wp-content/themes/lahma_maker/js/scripts.js#L86

gammaw:
@baxgas sure, the play events are still to be done, this is just a best effort traffic measurement (for the page visitors) so that we know more or less how many users we have after launch. I chose GA4 so that we can try the new technology (e.g., we'll have traffic data for the apps too). 

I know of the `head` config possibility (in `nuxt.config.js`) but there you cannot directly define code. I guess that's why they did this https://github.com/MatteoGabriele/vue-gtag little plugin but I didn't want do introduce this dependency (being old and the workaround seemed easy). 

What's wrong with the `app.html` solution?

gammaw:
We discussed off-line that it's OK to go with this solution now. I'm merging it.

PR #189, opened 2021-12-13
Reverse show-episode content in event

gammaw:
Show vs episode info tricked me again and I used the data in the opposite (wrong) way. Please review the fix and sorry for the extra PR. FYI @molnar-a


molnar-a (approved):
No probs :)

PR #190, opened 2021-12-13
#183 rare shows

baxgas:


pvj (approved):
Very very good workarounds for the unfortunate API situation, thanks for handling these rare show items well. Double thanks for @gammaw as the pairprogrammer on this :>

gammaw (approved):
thanks! latest... vars to be renamed later.

PR #192, opened 2021-12-14
#183 quick fixes after lunch

baxgas:


gammaw (approved):
Thanks for the quick fixes!

Issue #193, opened 2021-12-14
Hyperlinks not preserved when displaying WP data


gammaw:
E.g., Impressum page.

PR #196, opened 2021-12-14
#193 wp sanitize

baxgas:
@gammaw don't know which tags you need specifically, but I made a more generous filter for HTML sanitizer

### DIFF
--- a/app/components/news/Block.vue
+++ b/app/components/news/Block.vue
@@ -19,7 +19,7 @@
         </h5>
       </NuxtLink>
       <div class="news-text">
-        <div v-sanitize="truncatedNews" />
+        <div v-sanitize="[ sanitizeOptions, truncatedNews ]" />
       </div>
     </div>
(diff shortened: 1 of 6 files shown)


gammaw (approved):
Thanks! I wonder why it doesn't preserve thinkgs like `target="_blank"`? See for example links on the about us page (they should open in on a new tab). Do you have to allow the target attribute separately?

baxgas:
@gammaw yes, the sanitizer is a bit strict, but there are allow tags option of course, I think I usually put all the a tags to allow all, like here in Impressum https://github.com/mmmnmnm/lahmacun_rework/pull/203/commits/4d5280a0970b78c6a7fb1a4fe2835e626d995b4b#diff-169fab6db3c821a9491988943af38fb948520ea4cc1ab0bca88a916bf1dbe771
but which page is it exactly? from now I'll keep one pattern that works for all...

gammaw:
@baxgas it's Impressum here and About us in WP.

Issue #197, opened 2021-12-15
Fix up title, meta and header levels


baxgas:
h1 = Lahmacun radio,
h2 = page title
h3 = episode title
default og:image = logo (jpg with white background),
...
player changes the title of the page (might fix iPhone title)

PR #199, opened 2021-12-15
#197 metafixes

baxgas:
@gammaw I tidied the meta stuff a bit and also inserted the play triggering of the document title > very curious if it can solve the iPhone lock screen display!!! Microsoft Edge seems to use the page title when audio is playing on lock screen...


app/components/RadioPlayer.vue
```diff
@@ -368,17 +369,25 @@ export default {
...
+
       this.$store.commit('player/isArcsiPlaying', false)
       this.$store.commit('player/isStreamPlaying', true)
 
-      //Google Analytics 4 event: only send if it's a regular show on air
-      if(this.show_check){
+      // Google Analytics 4 event: only send if it's a regular show on air
+      if (this.show_check) {
         gtag('event', 'Radio play', {
-          'Show': this.show_title,
```
gammaw, line 377: @baxgas any reason why you removed the apostrophes? Are you sure it would work without them?

gammaw (commented):
@baxgas the changes look good, except one I didn't quite get (which seems unrelated with what the other changes are aiming at), can you please check (see inline comment)?

app/components/RadioPlayer.vue
baxgas: I think PHPstorm linter autofix might did it. It is ok without, sure it is a js object key, if it is a simple string it is correct like this

gammaw: Ok, let's see.

gammaw (approved):

PR #210, opened 2021-12-29
Use itemlist component

gygabor:


baxgas:
Thanks @gygabor 
I have a few remarks, 
- why is there no Show description on the Shows listing page? 
- you use the white border hover style on the Shows and News pages, but not on Archives, shouldn't it be the same everywhere? also the number of the items per row and the title/type style should be the same, no?
- I don't like that the white border hover sticks to the edge of the viewport (left-edge for the first, right-edge for the last in the row), there should be a bit (larger) margin around the container

gygabor:
> Thanks @gygabor I have a few remarks,
> 
> * why isn't there a Show description on the Shows listing page?
Because the descriptions on `ananasz` are also missing.
> * you use the white border hover style on the Shows and News pages, but not on Archives, shouldn't it be the same everywhere? also the number of the items per row and the title/type style should be the same, no?
Because I didn't update the archive page yet.
> * I don't like that the white border hover sticks to the edge of the viewport (left-edge for the first, right-edge for the last in the row), there should be a bit (larger) margin around the container
You're right. Something happened. I'll fix it.

baxgas:
There are show descriptions, but they apear on hover, replacing the cover
image 😊 no need to make it like that in your version, just use the same
list style everywhere as we agreed

Issue #213, opened 2022-01-06
First filter Arcsi episodes by number


baxgas:
(for the same day Shows, see NYE)

PR #214, opened 2022-01-06
#213 add correct sorting and prepare for latest supported node (15.14.0)

baxgas:


app/pages/shows/_slug/index.vue
```diff
@@ -132,6 +132,7 @@ export default {
         return showslist
           .filter(show => show.play_date < this.getToday)
           .filter(show => show.archived === true)
+          .sort((a, b) => new Date(b.number) - new Date(a.number))
           .sort((a, b) => new Date(b.play_date) - new Date(a.play_date))
```
gammaw, line 136: @baxgas if I understand correctly, this will first sort on the item ID and then on the play date? So for those items with the same date, the order will not change (after the first sort), but if the dates are different, the second sort may reshuffle the list, right?

gammaw (approved):
It seems to be an easy change, so I'll approve without trying it and waiting for the answer. @baxgas please check my question. Thanks!

app/pages/shows/_slug/index.vue
baxgas: @gammaw id cannot be changed, the number can, that was the reason of my choice, so if you eg. want to change the order after upload, you can by editing the number. the id is not flexible (what if you want to put a new item in the middle of the list)
the play_date should overwrite this, so the different date episodes are sorted after in the correct order

gammaw: Ok, thanks. I see. Number or ID, the principle is the same. Only it should be the opposite order, it seems :)

gammaw: > Only it should be the opposite order, it seems :)

Sorry, no. The higher number is in front, that's how we do it. All good.

baxgas:
@gammaw I made the latest first, as the date :-O should it be the other way for the time?

gammaw:
No, I think it's logical and good as it is for a default sorting. We can later introduce additional sorting options as we do on the current website.

baxgas:
Yesss, sorting options is a very good feature, hope we get to it soon! Should I open an issue with some basic dropdown choosing functionality, or should we wait for the UX/UI stuff?

gammaw:
This https://github.com/mmmnmnm/lahmacun_rework/issues/215 is it, no? :)

Issue #215, opened 2022-01-06
Add sorting change for Arcsi show Episodes list


baxgas:
![image](https://user-images.githubusercontent.com/36408851/148427439-05600433-0c7c-40ea-9c11-14f38719bc9b.png)

PR #216, opened 2022-01-16
#215 sort arcsi episodes

baxgas:


app/pages/shows/_slug/index.vue
```diff
@@ -44,8 +44,20 @@
...
+            <i v-if="airtimeAsc" class="fa fa-sort-numeric-desc" aria-hidden="true" />
+            <i v-else class="fa fa-sort-numeric-asc" aria-hidden="true" />
+            Order by Air time
+          </a>
+          <a id="alphabetical" ref="alphabetical" class="change-order-button" href="#" @click.prevent="sortAlphabeticaly">
+            <i v-if="alphabeticAsc" class="fa fa-sort-alpha-asc" aria-hidden="true" />
+            <i v-else class="fa fa-sort-alpha-desc" aria-hidden="true" />
+            Order by Title
+          </a>
+        </div>
```
gammaw, line 58: If this block (line 47-58) is the same as for the show subpage then you may have a mini component for that so that we don't duplicate code.

gammaw (approved):
Looks good! I saw it working on dev so I didn't build/test it myself. Thanks!
PS: I added an inline comment.

app/pages/shows/_slug/index.vue
baxgas: I think this is just a temporary quick-fix, don't know how the new UX/UI concept will handle this...

PR #218, opened 2022-01-16
#217 remove arcsi search

baxgas:


gammaw (approved):
Looks good! How about the News and Shows search field, we keep them? Are they not buggy? 

Disclaimer: I didn't build/test it myself.

Issue #198, opened 2021-12-15
Rich text for show description etc.


gammaw:
Root cause: arcsi stores plain text. 
Note for this issue: Probably there will be an arcsi side of this issue too but let's collect the Luca feedback here and then we distribute.

baxgas:
@pvj I think a `markdown editor` would be the [best](https://www.npmjs.com/package/@nuxtjs/markdownit), safety and simplicity-wise. hope it is available as a plugin for **jinja** – _what do you think?_

pvj:
The legacy now is HTML, so we just need to have HTML coming from the arcsi api show descriptions. I will check this. 

The long term editor for show hosts should be simple like markdown, I agree yes!

PR #219, opened 2022-01-16
#198 rich text allow for the show and episodes

baxgas:
I prepared the code for the html content of the show and episodes end-pages


app/components/item/Block.vue
```diff
@@ -17,6 +17,7 @@
       </h5>
       <div v-if="item.description" class="item-text mt-2">
         <div v-sanitize="item.description" />
+        Hello
```
gammaw, line 20: What's this?

app/components/ShowsLister.vue
```diff
@@ -15,7 +15,7 @@
           </div>
           <div class="show-desc p-6">
             <div class="showdescription">
-              {{ truncate(show.description, 200) }}
+              {{ show.description && truncate(stripHTMLTags(show.description), 200) }}
```
gammaw, line 18: What's the logical operator `&&` doing here? Does it check if `show.description` is empty? Can we use a more explicit notation instead, e.g., using `v-if` or the ternary operator?

gammaw (commented):
Thanks for the changes, looks good in general! I added some questions, can we check them before I approve?

app/components/item/Block.vue
baxgas: OOO, sorry, it was left there by accident :-(

app/components/ShowsLister.vue
baxgas: Yes, otherwise it would break if it tries to run the functions if show.description is not yet initialized
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Logical_AND

gammaw: Interesting semantics. Thanks for the explanation.

gammaw (approved):
Looks good now, thanks!

Issue #206, opened 2021-12-23
Display custom schedule info 


gammaw:
E.g., New Year's Eve live show, events etc. 

In WP it's easy, cause we simply edit a static HTML file, with our new site where we get the schedule info from arcsi, it's trickier. Any idea?

baxgas:
@gammaw yes, it is a bit trickier, but there could be some solutions, let's discuss which is the best...

gammaw:
Sure, why not try it here :) Maybe the easiest for now would be to add the one-time show in arcsi and set the regular shows temporarily inactive. What do you think?

baxgas:
One-day wordpress custom post with custom fields

PR #220, opened 2022-01-20
#206 custom schedule

baxgas:


app/components/schedule/Custom.vue
```diff
@@ -0,0 +1,146 @@
...
+      </div>
+      <div class="flex justify-between w-full schedule-infos">
+        <div class="onair-infos">
+          <div v-if="show.archive_lahmastore_base_url.includes(currentHost)">
+            <NuxtLink :to="show.archive_lahmastore_base_url.replace(currentHost, '')">
+              <b>{{ show.name }}</b>
+            </NuxtLink>
+          </div>
+          <div v-else-if="show.archive_lahmastore_base_url.includes('lahmacun.hu/shows/')">
+            <NuxtLink :to="'/shows/' + show.archive_lahmastore_base_url.substring(show.archive_lahmastore_base_url.lastIndexOf('/') + 1)">
```
gammaw, line 19: @baxgas this doesn't seem to work for the setting I added now in WP (which should be the actual setting for the upcoming custom event). The URL doesn't see to include the last part, I only have `localhost/shows` link in my local test. Can you please check?

app/components/schedule/Home.vue
```diff
@@ -67,9 +67,29 @@
...
             <img src="@/assets/img/arrow-schedule.svg" alt="" class="inline-block w-8 pb-1">
             {{ removeSeconds(show.end) }} -
-            <NuxtLink :to="'/shows/' + show.archive_lahmastore_base_url">
+            <div v-if="show.archive_lahmastore_base_url.includes(currentHost)" class="inline">
+              <NuxtLink :to="show.archive_lahmastore_base_url.replace(currentHost, '')">
+                <b>{{ show.name }}</b>
+              </NuxtLink>
+            </div>
+            <div v-else-if="show.archive_lahmastore_base_url.includes('lahmacun.hu/shows/')" class="inline">
+              <NuxtLink :to="'/shows/' + show.archive_lahmastore_base_url.substring(show.archive_lahmastore_base_url.lastIndexOf('/') + 1)">
```
gammaw, line 76: @baxgas same here (see above).

gammaw (commented):
Thanks, looks very nice and flexible! The URL part didn't work for me in my test, can you please check (see comment below)?

gammaw (approved):
So it was a `/` in the end of the URL (should be ommitted). Thanks!

Issue #208, opened 2021-12-26
Add correct documentation to the project (readme.md)


baxgas:
- [x] copy readme.md to the root
- [x] elaborate the setup, 
- [x] add infos about the scss watchers
- [x] add infos about the browser cors plugin

PR #222, opened 2022-01-22
#208 documentation

baxgas:
Move Readme.md to the root of the project and add more details about the setup


gammaw (approved):
Good addition! 

Comment: we could add here (or in a Wiki) the list of special pages in WP which are used by the website (such as custom schedule, rare events etc.). It's getting complex and now it's only in our head. Also, it would be nice to document the ACF structure that our code assumes, cause obviously the inside of WP is not publicly accessible. What do you think?

baxgas:
@gammaw yes, definitely, we should elaborate more. I put in some stuff that was noted by gygabor, but anything that could be some question for the community of future collaborators can go in, sure.

PR #224, opened 2022-01-22
#223 rare shows connect

baxgas:


gammaw (approved):
Such a nice PR! It not only adds a cool feature (and clear improvement), it also makes our code cleaner! Thanks!

gammaw:
@baxgas one comment: is there any chance to force in WP that _exactly one_ show is selected? The form now allows to select multiple shows for one rare show slot.

baxgas:
@gammaw I think they are quite separate entities (so they can be added/removed from the db via the admin), so they can't be connected easily :-/ 
The first checked is the one that is used in the frontend.

Issue #227, opened 2022-01-25
Periodically refresh the store state


baxgas:
Now it is only setting it on reloading the page, which can lead to stuck states, not refreshing shows etc.
What should be the trigger, time or some action? (should be on some idle time - maybe the clock that is in the header?)

PR #228, opened 2022-01-26
#227 rework store

baxgas:


gammaw (approved):
Thanks @baxgas for the optimisation and the custom/rare schedule fixes (nice usage of mapGetters)! I can imagine that some users never close the lahmacun website, for such users auto-"refresh" can be very useful. 

Disclaimer: I didn't build and tested these changes.

PR #230, opened 2022-01-26
#229 link uderline and Schedule styleing

baxgas:


app/components/item/Block.vue
```diff
@@ -44,7 +44,7 @@ export default {
     this.loadNewsImages()
   },
   methods: {
-    async loadNewsImages() {
+    loadNewsImages () {
```
gammaw, line 47: @baxgas why this change?

gammaw (approved):
Thanks for this too! 

Disclaimer: I didn't try it, only scanned through the changes and commented a question.

app/components/item/Block.vue
baxgas: doesn't need `async`, since it doesn't return a Promise to be resolved and doesn't have `await` for the variables init in the body, so it is not used as an async function. 
don't know why it was put in the first place like this

PR #231, opened 2022-01-31
#229 underline link

baxgas:
Sorry, the custom Schedule component didn't have the underlined links


gammaw (approved):
Thanks for finding and fixing it!

Issue #238, opened 2022-02-16
Use data refreshing only on client


baxgas:
Might cause node server overloads

PR #239, opened 2022-02-16
#238 setinterval refresh only on client

baxgas:


gammaw (approved):
Looks good! Disclaimer: I didn't try it, just looked at it :)

PS: I think you can always add all IT members as reviewers. We're all universal ;)

PR #242, opened 2022-02-22
#241 membership page

baxgas:


app/layouts/bare.vue
```diff
@@ -7,6 +7,6 @@
 
 <script>
 export default {
-  name: 'Bare'
+  name: 'BareLayout'
```
gammaw, line 10: @baxgas where you you use `BareLayout`? `error.vue` seems to use `bare` as a layout.

gammaw (approved):
Wow, that's a lot of stuff, for sure more than just adding the membership page ;)
@baxgas I checked the changes, looks good to me in general, but I didn't run the branch, so make sure use test it properly before merging! I added one comment where I wasn't sure.

app/layouts/bare.vue
baxgas: @gammaw  it is just some suggestion from the linter, doesn't influence the functionality

Issue #243, opened 2022-02-24
og:image metatag contains default value for news pages


gammaw:
This file seems to be the source of the bahavior: https://github.com/mmmnmnm/lahmacun_rework/blob/main/app/pages/news/_slug.vue

@baxgas can you please check?

gammaw:
I see that the following line should set the `og:image` value but for some reason it doesn't. 

https://github.com/mmmnmnm/lahmacun_rework/blob/3f8da2b0e5fc3b95c63772a1e70fe6ca9fc6b8ed/app/components/news/Full.vue#L81

PR #245, opened 2022-02-25
#243 og-image connect

baxgas:


app/pages/news/_slug.vue
```diff
@@ -67,6 +67,13 @@ export default {
       }
       const shortenMeta = this.stripHTMLTags(this.selectedNews?.excerpt?.rendered)
       return this.truncate(shortenMeta, 150)
+    },
+    metaOgImage () {
+      if (!this.selectedNews) {
+        return this.metaImage
+      }
+      const metaImageACF = this.selectedNews?.acf?.ogimage
+      return metaImageACF || this.metaImage
```
gammaw, line 76: This `||` notation in JS drives me crazy :)
Everytime I see it, I need to remind myself that it returns actual values and not a Boolean one. I guess it's different for Typescript.

gammaw (approved):
Thank! Looks good, I didn't try it. 

Note: until we don't source control the WP code (for ACF etc.), can we have a Wikipage or something where we keep a "change log"?

PR #257, opened 2022-04-08
Dev/add cookie banner#250

gammaw:
The banner code and the CMS content are ready to review. 

Highlights: 

- The banner is using the **Vuetify** framework (an UI extension for Vue apps), which is new in our code base. 
- The banner status is stored in the browser's **local storage** (to remember the decision - cookie-like behaivor).
- The banner **text is configurable in our CMS** -> the content is fetched via Wordpress API. 
- The cookie policy page has been created in our CMS too based on a paid Cookiebot service -> again, the content is fetched via Wordpress API. 
 
As a result, wrong wording etc. from the CMS should not block the PR. 

The code has been validated as yarn dev application and it has also been built and validated locally and on dev, so that a review may not need to run the code. Attached is a screenshot how it looks like. 

The PR is structured through commits and commit messages. 

Please review. 

image

### DIFF
--- a/app/assets/css/banner.css
+++ b/app/assets/css/banner.css
@@ -1,0 +1,3 @@
+.v-application--wrap{
+    min-height: 0vh;
+}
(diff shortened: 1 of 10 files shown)


tuz666 (approved):
Thanks for the descriptive commits, I tried to understand as much as I can but I could hardly say that I got everything. Despite that I tested it on my local env and it looked good to me.

PR #258, opened 2022-04-12
Add reload script to nginx container

gammaw:
Through a Let's Encrypt notification email I realised that the reload script in the nginx container was missing. I added it now to the the docker compose file. Please review.


pvj (approved):
Good catch! I wonder why and how I missed this.... S...orry...

Issue #261, opened 2022-04-27
Broken styling after cookie banner merge (side-effect)


gammaw:
The wrong styling (see example) sneaks in through a style override of `margin-right: 2 rem`. The root cause is unclear. Bug appeared after merging https://github.com/lahmacunradio/frontend/pull/257

![image](https://user-images.githubusercontent.com/28865986/165547337-327e707a-6b15-4f29-b4d6-7a823be6e5a3.png)

gammaw:
Probably the trouble is due to the following declaration which leaves out the tailwind styles (also in the assets dir). A quick is fix to add explicit styling the component definition. Tested locally. 

https://github.com/lahmacunradio/frontend/blob/84831f180104696f146b297a1a6910a30b97221c/app/nuxt.config.js#L112

PR #262, opened 2022-04-27
Add explicit styling

gammaw:
Quick fix for #261. Tested locally. Please review.


gammaw:
The isse needs further investigation.

Issue #264, opened 2022-04-30
Incorrect og:image link


gammaw:
Bug: URLs end with `/undefined`.

PR #268, opened 2022-05-09
Fix: Incorrect og:image link on episode subpage

gammaw:
It seems to be a timing issue with how fast the metadata is loaded based on the arcsi response. This best effort fix calls the code inline in Nuxt's `fetch()` instead of through nested subcalls. Note that metadata seems to be working well for other supages (e.g., show subpage), which use direct calls too. Tested locally and on dev for new- (slug) and old-style (ID) episode links. Please review.

#264

### DIFF
--- a/app/pages/shows/_slug/_id.vue
+++ b/app/pages/shows/_slug/_id.vue
@@ -126,9 +126,38 @@ export default {
       airtimeAsc: true
     }
   },
-  fetch () {
-    this.fetchEpisodeData()
-  },
+  async fetch () {
+      //Fetch episode data
+      await this.$axios.get(`${arcsiBaseURL}/show/${this.slug}/item/${this.id}`)
+        .then((res) => {
+          this.arcsiEpisode = res.data
+        })
+        .catch((error) => {
+          if (error.response.status === 404) {
+            //Use legacy API URL as fallback
+            this.$axios.get(`${arcsiBaseURL}/item/${this.id}`)
+              .then((res) => {
+                this.arcsiEpisode = res.data
+              })
+              .catch((error) => {
+                this.$sentry.captureException(new Error('Arcsi server not available ', error))
+                this.$nuxt.error({ statusCode: 404, message: 'Arcsi server not available' })
+              })
+          } else {
+            this.$sentry.captureException(new Error('Arcsi server not available ', error))
+            this.$nuxt.error({ statusCode: 404, message: 'Arcsi server not available' })
+          }
+        })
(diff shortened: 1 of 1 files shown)


app/pages/shows/_slug/_id.vue
```diff
@@ -126,9 +126,38 @@ export default {
       airtimeAsc: true
     }
   },
-  fetch () {
-    this.fetchEpisodeData()
-  },
+  async fetch () {
+      //Fetch episode data
+      await this.$axios.get(`${arcsiBaseURL}/show/${this.slug}/item/${this.id}`)
```
tuz666, line 131: Sorry for the stupid question but this and the URL at line 138 and 152 are the ones which we use to reach the Arcsi API, am I right?
I suppose these parameters (this.slug and this.id) in these URLs are route params, but I don't understand properly how and where from we get them. Could you explain that to me?

tuz666 (approved):
As I saw it was more like a refactoring of the current code, merging some smaller and similar functions into each other, not causing any problem to the logic.
I compared the prod and staging sites as well, I found everything works at the staging as it should.

app/pages/shows/_slug/_id.vue
gammaw: It's Nuxt logic and it comes from the path of the component, in this case `\app\pages\shows\_slug\_id.vue`, so basically these're template assets. Thanks for the review!

app/pages/shows/_slug/_id.vue
```diff
@@ -185,8 +214,7 @@ export default {
       return this.$moment(this.arcsiEpisode.play_date).format('yyyy. MMMM Do.')
     },
     episodeImage () {
-      const rootLink = mediaServerURL + this.slug + '/'
-      return rootLink + this.arcsiEpisode?.image_url
+      return this.arcsiEpisode?.image_url
```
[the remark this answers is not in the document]
gammaw: Oh I see. Thanks for pointing it out! But does it actually cause any trouble? I'm asking because the images seem to be displayed correctly on show subpages.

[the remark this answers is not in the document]
tuz666: It doesn't make any trouble, works as it should, just looks a bit ugly for my eyes 😅

app/pages/shows/_slug/_id.vue
tuz666: I should dive a bit deeper into these Nuxt stuff!

PR #269, opened 2022-05-10
Quick solution for bringing back stream image in radio player

gammaw:
Quick fix for bringing back stream image. Cherry picked of commits of @baxgas (see commit refs in the commit message). We agreed on a 2-step solution by first bringing back the image in exchange for the sand watch, then working out a more sensitive solution as needed. In this commit we have logo-play/pause-image-title (from left to right) instead of logo-play/pause-title-image (as we agreed) but I think it's good enough for the 1st step (we got this solution "for free" through old commits and I wouldn't want to tweak around it for a temporary solution). Tested locally and on dev (active there, check it out, here's a screenshot of it), please review. 

image

#225

### DIFF
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -498,12 +498,20 @@ export default {
         align-items: center;
         .now-playing-art {
           margin-right: 0.7rem;
+          @media (max-width: $extra-small-width) {
+            display: none;
+          }
         }
         .now-playing-main {
             flex: 1;
             min-width: 0;
             position: relative;
             height: 70px;
+            @media (max-width: $extra-small-width) {
+              height: auto;
+              margin-top: 0.25rem;
+              padding-right: 0.5rem;
+            }
             .media-body {
               overflow: auto;
             }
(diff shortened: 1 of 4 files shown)


molnar-a:
its only a minor thing, but between ~380px and ~400px the container doesn't shrink which might result in unhandled text overflow.
the first breakpoint below 380px handles this in a correct way.

if we can live with this, i'm happy to approve :)

app/components/RadioPlayer.vue
```diff
@@ -533,10 +539,8 @@ export default {
             text-overflow: ellipsis;
             overflow: hidden;
             white-space: nowrap;
```
gammaw, line 541: @molnar-a yeah I also noticed it in the browser emulator but it seems to render well on real devices. The key property is `white-space` here and in the code it's clear that it should be `nowrap` over 380px (this is what the simulator is doing too) but it was OK (wrapped) on all real devices and mobile browsers I checked. So yes, the code may not be ideal here but as long as it seems to work in the real world I wouldn't bother touching it. I suggest that you check on your phone too and if it's fine (wrapped) then we move on with this code for now.

molnar-a: fair enough! approved :)

molnar-a (approved):

PR #273, opened 2022-08-29
Adding Authentication-Tokens towards Arcsi requests

tuz666:
This change contains front-end side updates which needed for the support of latest Arcsi (our back-end server) enhancements.

In this [PR](https://github.com/lahmacunradio/arcsi/pull/153) we're about to introduce some security upgrades - like token based authentication on API endpoints or role based accesses on views - in our back-end server.

### DIFF
--- a/app/components/HeaderBlock.vue
+++ b/app/components/HeaderBlock.vue
@@ -23,7 +23,7 @@
 </template>
 
 <script>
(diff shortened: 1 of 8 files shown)


gammaw (approved):
@tuz666 thanks! The changes look good in principle. Haven't checked it for completeness (if all arcsi calls are covered), so please do a thorough manual validation on our dev servers. 

PS: Please next time add all IT wg members as reviewers.

PR #274, opened 2022-10-15
Dev/add auth token towards arcsi

tuz666:
This change contains front-end side updates which needed for the support of latest Arcsi (our back-end server) enhancements.

In this [PR](https://github.com/lahmacunradio/arcsi/pull/153) we're about to introduce some security upgrades - like token based authentication on API endpoints or role based accesses on views - in our back-end server.

### DIFF
--- a/app/components/HeaderBlock.vue
+++ b/app/components/HeaderBlock.vue
@@ -23,7 +23,7 @@
 </template>
 
 <script>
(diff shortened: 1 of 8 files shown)


gammaw (approved):
As I see contentwise it's the same as https://github.com/lahmacunradio/frontend/pull/273 so approve goes here too. Thanks!

PR #277, opened 2023-04-18
Add general purpose banner

gammaw:
Features: 

- Configurable (CMS-fed) content for message & button text
- Placed on top under the menu so that it doesn't conflict with arcsi player and cookie banner
- Colors: Lahmacun purple for background and yellow for button

Validation: 

- Deployed locally
- Deployed on dev and checked on mobile
- See screenshot

Disclaimer: 

- Current text is for demonstration only (it's not hard-coded, can be edited via CMS)

Credits:

- Main code by @baxgas (thanks man!), adjustments by me (see commits for detail)

Please review. 

image


tuz666 (approved):
Don't know too much about the frontend code, but I checked the dev page and found no problem with it.
Thanks for this PR!!

gammaw:
Thanks @tuz666 for the quick review & approve! Just added a minor commit to adjust the color code of the yellow we use (the same as for the cookie banner). Note: it's not deployed on the dev server. 

Please review again.

Issue #279, opened 2023-05-20
Get rid of code duplicate on schedule main page (schedule.vue) for group shows


gammaw:
Hint: use store data computed by the radio play component similar to schedule home.

Issue #278, opened 2023-05-09
Display external stream info


gammaw:
**Challenge**: frontend now displays current show content based on Azuracast's `nowplaying` API, but, in case of relayed streams, the API returns empty info and/or info filled by the stream's owner (not known a priori). 

**Solution candidate**: enhance current show display logic with [arcsi](https://github.com/lahmacunradio/arcsi)-based information (as a fallback, e.g., based on empty playlist in the API responce) and relying on CET-transformed browser time (see hints [here](https://linuxhint.com/convert-date-to-another-time-zone-using-javascript/) and [here](https://www.w3schools.com/jsref/jsref_tolocalestring.asp))*.

**Modules subject to change**: `RadioPlayer`, `schedule/Home` (aka ScheduleHome), `schedule` (and `schedule/FullItem` aka ScheduleFullItem). 

*Note that it seems that we display browser-local time no matter where we are: 

https://github.com/lahmacunradio/frontend/blob/a51452499f068bb442cdc42a9d356d221693919a/app/components/HeaderBlock.vue#L19

https://github.com/lahmacunradio/frontend/blob/a51452499f068bb442cdc42a9d356d221693919a/app/components/HeaderBlock.vue#L51-L52

PR #281, opened 2023-05-20
Dev/#278 display external stream info

gammaw:
Main changes: 

- Add displaying relayed streams feauture (see Wiki for usage: https://github.com/lahmacunradio/frontend/wiki/How-to-handle-relayed-streams-as-shows); 
   - Main logic: so far the player info (title, subtitle, image link) was solely filled based on the `nowplaying` API answer from the streaming server (Azuracast); this new feature introduces a fallback mechanism if such info is not available. Specifically, in case of stream relays there is no image link and the title and subtitle fields depend on the streamer. We use the combination of the `playlist` (empty) and `is_live` (false) fields to detect such a case and we fallback to compute the show based on the current CET time and arcsi. 
- Fix some CET issues (CET display and calculating CET times to be in sunch with arcsi schedules);
- Start using store (Vue's "global variable" concept) for sharing/reusing data across components (most notably in this PR: radio player component computes show info such as title, subtitle which is reused by, e.g., schedule on home page). 

Fixed issues:

- https://github.com/lahmacunradio/frontend/issues/278
- https://github.com/lahmacunradio/frontend/issues/280

Validation: 

- Tested locally with relay streams and for regression (live, scheduled shows etc.), 
- Deployed on dev server. 

Known issues (to be done later) and disclaimer: 

- https://github.com/lahmacunradio/frontend/issues/279
- Commits are sometimes chaotic, I needed to backtrack here and there. 
- Internal arcsi API (meant for arcsi UI) used to request unarchived show items too (see in usage doc above). 

**Please review until Monday if you can and sorry for the very short notice! It got to be a more complicated feature than I expected.**

### DIFF
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -127,6 +127,7 @@ import VueSlider from 'vue-slider-component/dist-css/vue-slider-component.umd.mi
 import 'vue-slider-component/dist-css/vue-slider-component.css'
 // import theme
 import 'vue-slider-component/theme/default.css'
+import { mapGetters } from 'vuex'
 
 export default {
   components: {
@@ -190,6 +191,9 @@ export default {
     }
   },
   computed: {
+    ...mapGetters({
+      todayShows: 'returnTodayShows'
+    }),
     streams () {
       const allStreams = []
       this.np.station.mounts.forEach(function (mount) {
@@ -243,7 +247,20 @@ export default {
       return (timeTotal) ? this.formatTime(timeTotal) : null
     },
     show_title () {
-      if (this.np.live.is_live) { return this.np.live.streamer_name } else { return this.np.now_playing.song.artist }
+      const title = ''
+      if (this.np.playlist !== '') // Show metadata can be served from Azuracast nowplaing API response
+        {
+          if (this.np.live.is_live) 
+            {title = this.np.live.streamer_name} else 
+            {title = this.np.now_playing.song.artist}
+        } else
+        {                         // Fallback: show metadata needs to be served from arcsi
+          title=''
+        } 
+        // Update show name in stream in store for other components
+        this.$store.commit('player/setStreamShowTitle', title)
+        return title
+
     },
     show_subtitle () {
       if (this.np.live.is_live) { return this.np.now_playing.song.title } else { return this.np.now_playing.song.title }
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -214,11 +214,7 @@ export default {
       return [...this.$store.getters.returnArcsiShows]
     },
     currentShowArcsi () {
-      if (this.np.live.is_live) { // live show
-        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.live.streamer_name))
-      } else { // pre-recorded show
-        return this.arcsiList.find(show => this.slugify(show.name) === this.slugify(this.np.now_playing.song.artist))
-      }
+      return this.arcsiList.find(show => this.slugify(show.name) === this.show_title)
     },
     time_percent () {
       const timePlayed = this.np.now_playing.elapsed
@@ -459,8 +455,6 @@ export default {
       this.$axios.get(this.nowPlayingUri).then((response) => {
         const npNew = response.data
         this.np = npNew
-        // Update show name in stream in store for other components
-        this.$store.commit('player/setStreamShowTitle', this.show_title())
         // Set a "default" current stream if none exists.
         if (this.current_stream.url === '' && npNew.station.listen_url !== '' && this.streams.length > 0) {
           let currentStream = null
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -255,7 +255,13 @@ export default {
             {title = this.np.now_playing.song.artist}
         } else
         {                         // Fallback: show metadata needs to be served from arcsi
-          title=''
+          for (let i = 0; i < this.todayShows.length; i++) {
+              let show = this.todayShows[i];
+              if (removeMinutesAndSeconds(show.start)<=getCurrentTimeHourCET()) { //as shows are timely ordered in input array, the first hit will be the currently running show
+                title=show.name;
+                break
+            }
+          }          
         } 
         // Update show name in stream in store for other components
         this.$store.commit('player/setStreamShowTitle', title)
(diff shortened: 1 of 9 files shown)


app/components/RadioPlayer.vue
```diff
@@ -242,14 +262,47 @@ export default {
...
+        } else // Fallback: show metadata needs to be served from arcsi
+        { title=this.show?.name } 
+        // Update show name in stream in store for other components
+        this.$store.commit('player/setStreamShowTitle', title)
+        return title
     },
     show_subtitle () {
-      if (this.np.live.is_live) { return this.np.now_playing.song.title } else { return this.np.now_playing.song.title }
+      let title = ''
+      if (this.nowPlayingInfoAvailable) // Show metadata can be served from Azuracast nowplaing API response
```
tuz666, line 283: typo (nowplaing -> now_playing)

app/components/RadioPlayer.vue
```diff
@@ -480,7 +570,33 @@ export default {
...
+        console.log(res.data)
+      })
+      .catch((error) => {
+        console.log(error)
+        this.$nuxt.error({ statusCode: 404, message: 'Show archive not found' })
+      })
+    },
+    // Helper method for getLatestEpisodeFromArcsi()
+    getLatestEpisode (episodes) {
+      const sortedItems = episodes.items
```
tuz666, line 589: Maybe using show instead of episodes would be more clear.

app/components/schedule/Fullitem.vue
```diff
@@ -94,61 +91,18 @@ export default {
...
       const descriptionFromArcsi = this.latestEpisodeData?.description
       return descriptionFromArcsi || this.show.description
     },
-    latestEpisodeImage () {
-      if (!this.latestEpisodeData) {
-        return false
-      }
-      return this.show.cover_image_url
-    },
     latestEpisodeTitle () {
```
tuz666, line 106: Shouldn't we rename this to latestArchivedEpisodeTitle, and latestEpisodeLink to latestArchivedEpisodeLink on L113?

app/components/schedule/Fullitem.vue
```diff
@@ -188,9 +145,7 @@ export default {
         })
     },
     getLatestEpisode (episodes) {
-      const sortedItems = episodes
-        .filter(show => show.play_date < this.getToday)
-        .filter(show => show.archived === true)
+      const sortedItems = episodes.items
         .sort((a, b) => b.number - a.number)
```
tuz666, line 149: using id instead of numbers here too?

app/plugins/mixinCommonMethods.js
```diff
@@ -104,6 +128,49 @@ export function getLanguageGraph (type) {
   }
 }
 
+// Input: all active arcsi shows
+// Output: shows grouped by days (every day is an array) and sorted tarting today's day (CET)
+export function groupShowsByDay (dbshows, rare_thu, rare_fri, customSchedule) {
```
tuz666, line 133: There is an [API](https://arcsi.lahmacun.hu/doc/#/Show%20Requests%20for%20frontend/get_show_schedule_by_day__day_) for querying the daily schedule, it's not handling the rare shows right now, but maybe we can use that for this use case.

app/store/player.js
tuz666, line 1: Nice work with the refactors :)

tuz666 (changes requested):
Thanks for the change, I found some typos and had some improvement ideas. ::)

app/components/RadioPlayer.vue
```diff
@@ -480,7 +570,33 @@ export default {
...
+      // We need the unarchived items too (no audio uploaded yet) to cover relay streams too
+      //.filter(show => show.archived === true) 
+      .sort((a, b) => b.number - a.number)
```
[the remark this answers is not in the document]
gammaw: You're right, I'm not sure we need it in the first place, given that we sort based on play date too. For now we can just assume these sortings give the same results. I'm a bit nervous now with last minute changing of the code.

app/components/RadioPlayer.vue
```diff
@@ -242,14 +262,47 @@ export default {
       const timeTotal = this.np.now_playing.duration
       return (timeTotal) ? this.formatTime(timeTotal) : null
     },
+    nowPlayingInfoAvailable (){
```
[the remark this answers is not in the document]
gammaw: I think we mostly use camelCase but it's not a formal convention, we can certainly set it up.

app/components/RadioPlayer.vue
```diff
@@ -186,10 +190,29 @@ export default {
...
       default_art_url: require('@/assets/img/stream/defaultshowart.jpg'),
       default_azuracast_art_url: require('@/assets/img/stream/generic_song.jpg'),
-      docTitleSetter: null
+      docTitleSetter: null,
     }
   },
   computed: {
+    ...mapGetters({
+      rareShows: 'returnRareShows',
+      customSchedule: 'returnCustomSchedule',
```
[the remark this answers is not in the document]
gammaw: Yeah, that should be unified. At some point I started using the direct access (without declared getters) where no legacy dependancies (using getter names in the code elsewhere) needed declared getters. Future work, I would say.

app/components/schedule/Fullitem.vue
gammaw: You're right, now that we only request unarchived items. can we handle it as future work? I wouldn't want to risk refactoring now.

app/components/schedule/Fullitem.vue
gammaw: See my above comment.

app/plugins/mixinCommonMethods.js
gammaw: Absolutely, we could save a lot of computation. Future work?

gammaw:
Thanks @tuz666! As a best effort and last minute reaction, I pushed a minimal change and replied on each of the comment. Please check.

app/components/RadioPlayer.vue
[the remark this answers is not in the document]
tuz666: Yepp, it could be a really rare case, go on with the already tested solution.

app/components/RadioPlayer.vue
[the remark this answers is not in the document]
tuz666: Okay, go with the current state then.

app/components/schedule/Fullitem.vue
tuz666: Yepp, I can create issues/cards for them.

app/plugins/mixinCommonMethods.js
tuz666: deal

tuz666 (approved):

Issue #289, opened 2023-06-20
Don't list future episodes on public pages


gammaw:
See for example: ![image](https://github.com/lahmacunradio/frontend/assets/28865986/924470e8-4ce0-42d9-819d-0f39e19d8e9b)

Related: https://github.com/lahmacunradio/frontend/issues/288

Issue #288, opened 2023-06-11
Use dedicated arcsi API to query non-archived items


gammaw:

tuz666:
Zero step is merged on Arcsi side, '/show/<show_slug>/page' now has now two filtering options ('archived' and 'latest')
As a [workaround](https://github.com/lahmacunradio/arcsi/blob/master/arcsi/api/show.py#L398) for the current usage 'archived' is always true for requests coming from the frontend.
Frontend change should be merged which adds the 'archived' filter for the current plain API call.
After that we can remove the temporary workaround in the Arcsi code and we can change the 'show/<id>' API to the unfiltered '/show/<show_slug>/page' request.
FYI @gammaw

gammaw:
- [ ] Frontend update (all except show/item pages): https://github.com/lahmacunradio/frontend/tree/dev/%23289-dont-list-future-eps
- [ ] Frontend deploy (of above branch) after suitable review (https://github.com/lahmacunradio/frontend/pull/291)
- [ ] Arcsi adjustment: https://github.com/lahmacunradio/arcsi/pull/175 & deploy
- [ ] Frontend update (show/item pages) here https://github.com/lahmacunradio/frontend/tree/dev/%23289-dont-list-future-eps (same branch as above) & review here https://github.com/lahmacunradio/arcsi/pull/175 & merge/deploy

FYI @tuz666

PR #291, opened 2023-06-21
Dev/#289 don't list future eps

gammaw:
This is primarily a bug fix change and also an arcsi API usage consolidation one (thanks @tuz666 & @pvj for the great work on the new arcsi API version!). As it's bug fix, please review asap. 

Related issues:
- #288 
- #289 

Tests/validation done: 
- Deployed locally; 
- Local regression tests (quick & best effort): custom schedule, live show, relayed show, regular schedule; 
- Deployed on dev (for your quick validation);

### DIFF
--- a/app/components/RadioPlayer.vue
+++ b/app/components/RadioPlayer.vue
@@ -566,31 +566,15 @@ export default {
       this.streamModal = false
     },
     getLatestEpisodeFromArcsi() {
-    // TODO: use dedicated frontend API instead of one meant for arcsi UI (rationale: we need the unarchived episodes too for relay logic and there was no such frontend arcsi API available)  
-    // TODO: see also related workaround in getLatestEpisode ()
-    this.$axios.get(arcsiBaseURL + '/show/' + this.show.id, config)
+    this.$axios.get(arcsiBaseURL + '/show/' + this.slugify(this.show.name) + '/page?filter=latest', config)
       .then((res) => {
-        this.latestEpisodeData = this.getLatestEpisode(res.data)
+        this.latestEpisodeData = res.data.items //note that it's a sinlge item with filter latest, plural in var name is misleading
       })
       .catch((error) => {
         console.log(error)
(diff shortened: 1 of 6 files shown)


app/components/RadioPlayer.vue
```diff
@@ -566,31 +566,15 @@ export default {
       this.streamModal = false
     },
     getLatestEpisodeFromArcsi() {
-    // TODO: use dedicated frontend API instead of one meant for arcsi UI (rationale: we need the unarchived episodes too for relay logic and there was no such frontend arcsi API available)  
-    // TODO: see also related workaround in getLatestEpisode ()
-    this.$axios.get(arcsiBaseURL + '/show/' + this.show.id, config)
+    this.$axios.get(arcsiBaseURL + '/show/' + this.slugify(this.show.name) + '/page?filter=latest', config)
```
tuz666, line 569: Here we want to return the latest `item` for all the episodes of the `show` (including not yet aired episodes), right?
Just because there is a workaround in the Arcsi right now which adds the `archived` flag as "true" for all the requests coming from the frontend (checks if the origin URL contains `https://www.lahmacun.hu`)

tuz666 (commented):
Thanks for the quick PR @gammaw !
Most of the stuff looks good for me, I have only one concern at the `RadioPlayer.vue` component.
I think we should do that change in a next PR, because the future episodes won't be returned with the current Arcsi workaround.
What do you think about this?

app/components/RadioPlayer.vue
gammaw: Ok, reverted my original commit on this file. @tuz666 please check.

tuz666 (dismissed):
Nice move with that revert, from my side it looks good now :)

gammaw:
@tuz666 I reverted the revert as set out in https://github.com/lahmacunradio/frontend/issues/288. Please review (only changes in RadioPlayer.vue).

tuz666 (approved):
Thanks for the "fix"! Looks good for me :)

Issue #292, opened 2023-06-21
Migrate from UA to GA4


gammaw:

gammaw:
See PR https://github.com/lahmacunradio/frontend/pull/293 for details on how analytics is done on GA4 side.

PR #293, opened 2023-06-27
Dev/#292 migrate from ua to ga4

gammaw:
**Rationale**: in 4 days, Universal Analytics (UA) will stop to deliver data. Our show and episode titles are currently sent via UA custom events. As a result, starting in 4 days, we won't have this information. 

**This PR**: changes the measurement ID from UA to GA4 (as configured in the Google Analytics admin page) and sends GA4-compatible custom events. These custom events are different from our prev. UA ones: 
- Instead of using the built-in `event category` and `event action` parameters, we now introduced custom `show_title` and `episode_title` parameters (note that we don't use Google Tag Manager).  
- These new parameters are handled (in suitable queries, specifically, "explorations" and "custom dimensions") in the Google Analytics admin. 

**Disclaimer and tests**:
-  Measurement ID could be taken from local file but as it needs to be in the page's code (as a parameter to the Google JS file), it's not a secret, and this is how we did it so far too. 
- Custom events have been tested on dev, and we can reproduce all stats that we've had so far. 

**Related issues**:
- https://github.com/lahmacunradio/frontend/issues/292

Please review asap (see above timeline).

### DIFF
--- a/app/app.html
+++ b/app/app.html
@@ -3,7 +3,7 @@
   <head {{ HEAD_ATTRS }}>
     {{ HEAD }}
     <!-- Global site tag (gtag.js) - Google Analytics -->
-    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-149605744-1"></script>
+    <script async src="https://www.googletagmanager.com/gtag/js?id=G-333552546"></script>
     <script>
       window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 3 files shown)


tuz666 (approved):
Thanks for the quick fix and the validation. Looks good to me :)

Issue #294, opened 2023-08-10
Show tags


gammaw:

tuz666:
We discussed with @baxgas and @gammaw that first we have to update the Swagger APIs to return the tag's `clean_name` fields everywhere where it is needed.
After this the frontend will be able to use this field for the `/tags/<clean_name>` query.

gammaw:
Arcsi side update: https://github.com/lahmacunradio/arcsi/pull/179

Update is not yet deployed. 

FYI @tuz666 @baxgas

gammaw:
@baxgas some final (hopefully smaller) change requests: 

- Route of tag pages: lahmacun.hu/tag/[tag_name]
- Title (header) of tag pages: TAGGED BY [TAG_NAME]
- On a tag page: list all tags of every show and episode (und the ep/show's "card")

gammaw:
@baxgas two more last last minute change request but tell us if it's too much for this project. Hopefully smaller changes: 

- Separate tags for shows and episodes on tag cloud page. 
- On Home: limit tags for episodes for max 5 tags.

gammaw:
The corresponding dev branch: https://github.com/lahmacunradio/frontend/issues/294

PR #298, opened 2023-09-06
#294 show tags

baxgas:


app/pages/shows/index.vue
gammaw, line 1: I thought that we add tags for the shows here too. But it's ok not to, cause one might think that we also search in tags (which I don't think we do in the current version). FYI @tuz666

gammaw (approved):
Thanks & approval! ;)

Issue #296, opened 2023-08-28
Maiden show renders error on schedule page


gammaw:
**Bug**: no episode data available for maiden show (see pic)

**Fix(es)**: 

- Robust using of API on frontend side and/or
- Adjustment on arcsi side so that meaningful info is returned if there is no data

FYI @tuz666 

![image](https://github.com/lahmacunradio/frontend/assets/28865986/f87a1a00-9fc6-43d3-b752-e8375579b705)

gammaw:
This is how the render looks like after above quick fix: 

![image](https://github.com/lahmacunradio/frontend/assets/28865986/5f1cbf98-d456-4b03-b140-eb19252286be)

PR #299, opened 2023-09-13
Bug fix: quick frontend side fix for schedule corner case w/ no episode

gammaw:
This is a frontend side quick fix so that error is not generated when the API calls fails. Note that the render is OK (robust) as the episode data is simply not shown.  #296


tuz666 (approved):
It was tested previously and was deployed on the prod, looks good from my side!

Issue #300, opened 2023-09-17
Sentry error (getPropagationContext is not a function)


gammaw:
Error shown with https://github.com/lahmacunradio/frontend/commit/095df60a812a28be701fc494ff5db418599a3818

More details: ![image](https://github.com/lahmacunradio/frontend/assets/28865986/4418a67a-a51b-40b2-8d50-2e35ea493e23)

gammaw:
Hint: https://github.com/getsentry/sentry-javascript/issues/8538

Thanks @tuz666

PR #301, opened 2023-09-21
Bug fix: Sentry

tuz666:
During the latest Tag related change we had to update Node.
Most probably this Node version bump caused a compatibility issue with the previously working and well configured Sentry lib.
The failure was recorded in #300 and it was resolved temporarily by the manual removal of Sentry.
This change contains a Sentry lib update and the removal of some unnecessary lines from its config.
The changes were tested on the dev server and after that I also verified it on the prod.


app/nuxt.config.js
```diff
@@ -72,11 +72,6 @@ export default {
   },
   sentry: {
     dsn: process.env.SENTRY_DSN,
-    publish_release: true,
```
gammaw, line 75: Do we need this removal for the fix to work? Any idea what these lines exactly do?

app/package.json
```diff
@@ -15,8 +15,8 @@
     "@babel/core": "^7.0.0-0",
     "@babel/plugin-proposal-optional-chaining": "^7.14.5",
     "@nuxtjs/axios": "^5.13.6",
-    "@nuxtjs/sentry": "^5.1.5",
-    "@sentry/tracing": "^7.62.0",
```
gammaw, line 19: Any idea what this is and why it needs to be removed?

app/package.json
```diff
@@ -15,8 +15,8 @@
...
+    "@nuxtjs/sentry": "^7.5.0",
+    "@sentry/webpack-plugin": "^1.20.1",
```
gammaw, line 19: Why is this needed?

gammaw (commented):
Thanks @tuz666! Allow me some questions before we move on.

app/package.json
tuz666: @sentry/tracing is deprecated and no longer needed from @nuxtjs/sentry v7.1.12
https://github.com/nuxt-community/sentry-module/releases/tag/v7.1.12

app/nuxt.config.js
tuz666: This configuration would be used to monitor different releases, but it was wrongly configured (publishRelease instead of publish_release), during my investigation I tried to set this up properly but unfortunately I couldn't make it work.

app/package.json
tuz666: The @nuxtjs/sentry version bump solved our issue.
@sentry/webpack-plugin package could be used for the publishRelease function if we want to implement it in the future.
https://sentry.nuxtjs.org/guide/releases

gammaw (approved):
Thanks @tuz666 for the detailed answers!

PR #307, opened 2024-01-07
Dev/#306 env load

baxgas:


baxgas:
@gammaw we need to install this dotenv package because the .env variable is loaded on top level because it is inserted in the auth of the APi fetch

gammaw (approved):
Let's do it, sorry for the delay.

PR #308, opened 2024-01-27
Dev/#305 stripe donations

baxgas:


app/pages/donation-stripe.vue
```diff
@@ -0,0 +1,132 @@
...
+        <div class="mb-4">
+          <h2>{{ donationContent.title.rendered }}</h2>
+        </div>
+        <div class="grid md:grid-cols-2 md:gap-8 gap-4">
+          <div>
+            <div v-sanitize="[sanitizeOptions, donationContent.content.rendered]" />
+          </div>
+
+          <div>
+            <form action="https://cms.lahmacun.hu/wp-json/stripe/return_checkout_session_listener" method="GET">
```
gammaw, line 19: This should not be hard-wired cause we cannot test it.

app/constants/index.js
```diff
@@ -51,3 +46,14 @@ export const mediaURL = contentApiURL + '/media'
...
 export const customScheduleURL = contentApiURL + '/pages/3679'
+
+export const donateURL = contentApiURL + '/pages/2'
+export const membershipURL = contentApiURL + '/pages/1425'
+
+export const membershipStripeURL = contentApiURL + '/pages/4980'
+export const membershipStripeCancelURL = contentApiURL + '/pages/4992'
+export const membershipStripeThanksURL = contentApiURL + '/pages/4994'
+
+export const donateStripeURL = contentApiURL + '/pages/5088'
```
gammaw, line 57: Can we refer to these pages via the page **title** in the URL? Like this it's hardly testable (we cannot guarantee that we have the same page number on the dev server).

gammaw (changes requested):
See inline comments.

app/constants/index.js
gammaw: Never mind! Page ID-s are editable in WP!

app/pages/donation-stripe.vue
```diff
@@ -0,0 +1,132 @@
...
+              <div>
+                <p class="mb-2">
+                  <strong>
+                    You can choose from the following options:
+                  </strong>
+                </p>
+                <div class="my-2">
+                  <input type="radio" id="option1" name="is_recurring" value="no" checked />
```
gammaw, line 27: Could we style it a bit more Lahma-like? The blue seems to me mismatchy

gammaw (commented):
Left another comment on the radio button styling

gammaw:
In addition, please: 

- In https://github.com/lahmacunradio/frontend/blob/dev/%23305-stripe-donations/app/.env.template only keep `SENTRY_DSN`, `SUBSCRIPTION_CANCEL_URL`, `DONATION_STRIPE_FORM_URL`, `MEMBERSHIP_STRIPE_FORM_URL` and make sure that the others are not used in code. 
- Change the radio button style to more modern style (in synch with fancy new membership drop-down list).

app/pages/donation-stripe.vue
baxgas: OK, good idea!

app/pages/membership-stripe.vue
```diff
@@ -106,13 +104,14 @@ export default {
...
           !(show.archive_lahmastore_base_url === 'off-air' || !show.active)
-        )).sort((a, b) => a.name.localeCompare(b.name))
+        )).sort((a, b) => a.name.localeCompare(b.name)).map(show => show.name)
       }
       return null
     },
   },
   methods: {
     selectShow(showname) {
+      console.log(showname, 'selcted show')
```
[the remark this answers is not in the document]
baxgas: thanks, I removed the console.log

app/pages/membership-stripe.vue
```diff
@@ -16,27 +16,25 @@
...
+            <form :action="$config.membershipStripeFormUrl" method="GET">
+              <div class="selector">
+
+                <v-app>
+                  <v-container>
+                    <v-select label="Select your show" name="show_name" v-model="show_name"
+                      :items="arcsiShowsList"></v-select>
+                  </v-container>
+                </v-app>
+                <input type="hidden" name="show_name">
```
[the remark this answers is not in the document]
baxgas: yes, is the url correct? because the selector works fine. I added a fix also for the value now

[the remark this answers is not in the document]
tuz666: Yes, now I realized, that there was also an issue with the Stripe secret key and the configured URL on the CMS/"backend" side.
I resolved that misconfiguration and it works now, but this duplicated `show_name` parameter still looks a bit odd for me, even both of them has the same value.
I tested it manually on the dev environment and if you remove [this line](https://github.com/lahmacunradio/frontend/pull/308/files#) it will work fine.

tuz666 (changes requested):
Except those two comments of mine, I have one more, shouldn't we use `donation` as prefix everywhere (pages, variables, constants) instead of the various `donate` and `donate` prefixes?
I think it would improve the readability.

app/pages/donation-stripe.vue
```diff
@@ -0,0 +1,132 @@
...
+  @apply block p-2 my-2 border w-full text-gray-500 z-10 relative;
+}
+
+.show-select:focus {
+  @apply outline-none;
+}
+
+.show-select.showSelected {
+  @apply text-black;
+}
```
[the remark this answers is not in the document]
baxgas: Thank you for finding it, it was a legacy code, I removed it now

app/pages/membership-stripe.vue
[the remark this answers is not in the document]
baxgas: Oh, OK, I thought the v-model was only making the local state and not the form params, thanks, fixed it now

app/pages/donate.vue
```diff
@@ -1,27 +1,65 @@
...
+
         </div>
-      </article>
+      </div>
     </div>
   </div>
 </template>
 
 <script>
-import { donateURL } from '~/constants'
```
tuz666, line 19: can we remove this `donateURL` from the `'~/constants'`?
or is it used anywhere else?

app/pages/membership.vue
```diff
@@ -5,22 +5,52 @@
...
+          </div>
+
+        </div>
+      </div>
     </div>
   </div>
 </template>
 
 <script>
-import { membershipURL } from '~/constants'
```
tuz666, line 19: same for `membershipURL`?

tuz666 (changes requested):

tuz666 (approved):
It looks good and everything was tested on the dev server.

gammaw (approved):
Thanks for the good work @baxgas @tuz666 !

PR #309, opened 2024-02-21
fix(*): fix positioning

baxgas:


gammaw (approved):
Like we discussed, it's a quick fix but we'll need to get rid of Vuetify.

PR #312, opened 2024-03-11
Quick fix for donate and membership parameters

tuz666:
- Remove additional `donate` param from Donate page's Stripe params
- Add hidden `show_name` param for Membership page's Stripe param


gammaw (approved):
Code looks good and validation tests passed, thanks for the quickness @tuz666!

PR #319, opened 2024-08-20
Dev/#316 add member and sponsor logos to footer

baxgas:


gammaw (approved):
Very elegant solution! Thanks!

gammaw (approved):
Elegant simplistic solution! Grats & thanks!

PR #326, opened 2026-07-02
Remove custom schedule and rare shows

tuz666:
There are multiple reasons why we should get rid of the custom schedule and rare show workarounds:
- The first and most important is that the mobile app is not supporting them
- The custom schedule was pretty convenient if we wanted to override only one day, but there were multiple occassions when we had to overwrite multiple, consecutive days, which was not possible
- Another reason why we used custom schedule was because we didn't have to change the `active` field for each overriden shows, but now it can be edited by the show hosts
- As soon as we introduce the new a/b/c/d schedule the rare shows will also lose their importance


pvj (dismissed):
Only checked on the surface. We should give it a go!

pvj (approved):
I don't know what my stale review comments were. It looks good to go again!

